PR #1, opened 2021-08-03
Add Vpc resource

brycahta:
Issue: [#489](https://github.com/aws-controllers-k8s/community/issues/489)

* Adding create/delete VPC functionality with smoke tests
* Adding Makefile
* Adding metadata file


apis/v1alpha1/vpc.go
```diff
@@ -0,0 +1,127 @@
...
+	// The IPv4 network range for the VPC, in CIDR notation. For example, 10.0.0.0/16.
+	// We modify the specified CIDR block to its canonical form; for example, if
+	// you specify 100.68.0.18/18, we modify it to 100.68.0.0/18.
+	// +kubebuilder:validation:Required
+	CIDRBlock *string `json:"cidrBlock"`
+	// Checks whether you have the required permissions for the action, without
+	// actually making the request, and provides an error response. If you have
+	// the required permissions, the error response is DryRunOperation. Otherwise,
+	// it is UnauthorizedOperation.
+	DryRun *bool `json:"dryRun,omitempty"`
```
RedbackThomson, line 40: Please add to the ignored fields list

pkg/resource/vpc/hooks.go
```diff
@@ -0,0 +1,37 @@
...
+import (
+	"errors"
+
+	svcsdk "github.com/aws/aws-sdk-go/service/ec2"
+)
+
+// If any required fields in the input shape are missing, AWS resource is
+// not created yet. Return False to indicate to callers that the
+// resource is not yet created.
+func isRequiredFieldsMissingFromInput(r *resource) bool {
```
RedbackThomson, line 25: For godoc, please use the name of the method as the first word in the doc. e.g.:
```
isRequiredFieldsMissingFromInput returns True if any required fields in the input shape are missing...
```

pkg/resource/vpc/hooks.go
```diff
@@ -0,0 +1,37 @@
...
+}
+
+// Add requested resource VpcId to ListRequest. Return error to indicate to callers that the
+// resource is not yet created.
+func addIdToListRequest(r *resource, input *svcsdk.DescribeVpcsInput) error {
+	if r.ko.Status.VPCID != nil {
+		input.VpcIds = []*string{r.ko.Status.VPCID}
+		return nil
+	}
+	return errors.New("unable to extract vpcId from Kubernetes resource")
```
RedbackThomson, line 36: super nit: Early return pattern

```suggestion
	if r.ko.Status.VPCID == nil {
		return errors.New("unable to extract VPCID from Kubernetes resource")
	}
	input.VpcIds = []*string{r.ko.Status.VPCID}
	return nil
```

templates/hooks/vpc/sdk_read_many_post_build_request.go.tpl
```diff
@@ -0,0 +1,4 @@
+	err = addIdToListRequest(r, input)
+	if err != nil {
+		return nil, ackerr.NotFound
+	}
```
RedbackThomson, line 4: nit: Conditional can be shortened to a one-liner
```suggestion
	if err = addIdToListRequest(r, input); err != nil {
		return nil, ackerr.NotFound
	} 
```

test/e2e/tests/test_vpc.py
```diff
@@ -0,0 +1,97 @@
...
+            if vpc["VpcId"] == vpc_id:
+                return vpc
+
+        return None
+
+    def vpc_exists(self, ec2_client, vpc_id: str) -> bool:
+        return self.get_vpc(ec2_client, vpc_id) is not None
+
+    def test_smoke(self, ec2_client):
+        resource_name = "vpc-ack-test"
```
RedbackThomson, line 57: Please use the [`resources.random_suffix_name`](https://github.com/aws-controllers-k8s/test-infra/blob/main/src/acktest/resources.py#L54) method to autogenerate a name. Otherwise if multiple test runs are invoked in parallel they could collide.

RedbackThomson (changes requested):
Super clean. Seems fairly straightforward and I hope will be a good basis for a lot of the other resources. 

Biggest thing is just to remove the `DryRun` field from spec - add to list of ignored fields.

I've left some minor nits about code style, but nothing major from me otherwise.

vijtrip2:
Looks solid beside incorporating Nick's comments.

pkg/resource/vpc/hooks.go
```diff
@@ -0,0 +1,37 @@
...
+// isRequiredFieldsMissingFromInput returns True if any required
+// fields in the input shape are missing meaning AWS resource is
+// not created yet.
+func isRequiredFieldsMissingFromInput(r *resource) bool {
+	return r.ko.Status.State == nil
+}
+
+// addIdToListRequest adds requested-resource VpcId to ListRequest. Return error to indicate to callers that the
+// resource is not yet created.
+func addIdToListRequest(r *resource, input *svcsdk.DescribeVpcsInput) error {
```
RedbackThomson, line 31: super nit: `Id` -> `ID`. We try to use the proper capitalisation in all the Go code except for when it's an external type.

RedbackThomson (approved):
Great! Will let someone else check it out, but looks solid to me.

A-Hilaly:
Neat :+1:
/ok-to-test

A-Hilaly:
holding until we add pre-submit jobs for ec2-controller (https://github.com/aws-controllers-k8s/test-infra/pull/82)
/hold

apis/v1alpha1/types.go
```diff
@@ -27,3 +27,3017 @@ var (
...
+	Tags                       []*Tag    `json:"tags,omitempty"`
+	TransitGatewayAttachmentID *string   `json:"transitGatewayAttachmentID,omitempty"`
+	TransitGatewayID           *string   `json:"transitGatewayID,omitempty"`
+	VPCID                      *string   `json:"vpcID,omitempty"`
+	VPCOwnerID                 *string   `json:"vpcOwnerID,omitempty"`
+}
+
+// The VPN tunnel options.
+type TunnelOption struct {
+	DpdTimeoutAction     *string `json:"dpdTimeoutAction,omitempty"`
```
jaypipes, line 2774: @brycahta "Dpd" here should be normalized to "DPD". If you'd like to submit a patch to the code-generator repo correcting this, @RedbackThomson or @A-Hilaly can walk you through the process (it's a relatively simple change to https://github.com/aws-controllers-k8s/code-generator/blob/main/pkg/names/names.go)

apis/v1alpha1/types.go
```diff
@@ -27,3 +27,3017 @@ var (
...
+	Code        *string `json:"code,omitempty"`
+	Description *string `json:"description,omitempty"`
+	EventID     *string `json:"eventID,omitempty"`
+	EventType   *string `json:"eventType,omitempty"`
+}
+
+// Information about the instances to which the volume is attached.
+type VolumeStatusAttachmentStatus struct {
+	InstanceID    *string `json:"instanceID,omitempty"`
+	IoPerformance *string `json:"ioPerformance,omitempty"`
```
jaypipes, line 3022: Likewise this should be normalized to "IOPerformance" (in a future PR to the code-generator, nothing to do for this PR)

generator.yaml
```diff
@@ -1,4 +1,6 @@
 ignore:
+  field_paths:
+    - CreateVpcInput.DryRun
```
jaypipes, line 3: :+1:

pkg/resource/vpc/hooks.go
```diff
@@ -0,0 +1,37 @@
...
+package vpc
+
+import (
+	"errors"
+
+	svcsdk "github.com/aws/aws-sdk-go/service/ec2"
+)
+
+// isRequiredFieldsMissingFromInput returns True if any required
+// fields in the input shape are missing meaning AWS resource is
```
jaypipes, line 23: Couple things...

1) Maybe it was a copy-pasta mistake, but this function isn't analyzing "the input shape". It's analyzing the CR (`r *resource`) to check if `Status.State` is nil.

2) This is actually what the code-generator's [`pkg/generate/code.CheckRequiredFieldsMissingFromShape` generator function](https://github.com/aws-controllers-k8s/code-generator/blob/843e3dd57998a61a0cd460de4da6ea1467a6f036/pkg/generate/code/check.go#L60-L93) should be used for. As you can see from that code, however, we don't yet implement it for ReadMany operations. We should get this done and you won't need a custom hook for this after that.

pkg/resource/vpc/hooks.go
```diff
@@ -0,0 +1,37 @@
...
+// not created yet.
+func isRequiredFieldsMissingFromInput(r *resource) bool {
+	return r.ko.Status.State == nil
+}
+
+// addIDToListRequest adds requested-resource VpcId to ListRequest. Return error to indicate to callers that the
+// resource is not yet created.
+func addIDToListRequest(r *resource, input *svcsdk.DescribeVpcsInput) error {
```
jaypipes, line 31: And this function should not be necessary if the code-generator's [`pkg/generate/code.SetSDK` generator function](https://github.com/aws-controllers-k8s/code-generator/blob/843e3dd57998a61a0cd460de4da6ea1467a6f036/pkg/generate/code/set_sdk.go#L75) understood that for ReadMany operations, if it finds a field on the input shape that is the plural of an identifier field (in this case, VPCID -> VpcIds) then it should populate the input shape for the ReadMany operation with a slice of one element containing the resource's identifier field.

We can work on adding this functionality (we need it for S3 as well)

jaypipes (approved):
Nice work on this, @brycahta! Left some comments inline that are mostly to the ACK core team (/cc @vijtrip2 @RedbackThomson @A-Hilaly) for things we need to do to the code-generator to avoid the need for some of the custom code you've had to place in here.

pkg/resource/vpc/hooks.go
brycahta: good call too much 🍝

apis/v1alpha1/types.go
brycahta: when you say normalized does that mean it should be aligned with go styling?

jaypipes: Yeah, initialisms should be capitalized fully when exported and lowercases fully when not exported according to Go style guidelines: https://github.com/golang/go/wiki/CodeReviewComments#initialisms

RedbackThomson:
Pod initialising timed out? Weird

/retest ec2-kind-e2e

RedbackThomson:
Pod initialising timed out? Weird

/retest

RedbackThomson:
Added the EC2 test role
/retest

jaypipes (approved):
Nice. :+1:

PR #6, opened 2021-09-03
Adding Route table

brycahta:
Issue #, if available: [#489](https://github.com/aws-controllers-k8s/community/issues/489)

Description of changes:
* adds route table resource
* refactors e2e tests to share common resources (i.e. VPC)
  * otherwise, vpcs would need to be created for each test which is inefficient and may exceed quota
  * now, only 2 vpcs are created per test run
  * could not use `@pytest.fixture(scope="session")` on vpc because we use [pytest-xdist](https://github.com/pytest-dev/pytest-xdist/issues/271)
* adds route table tests


test/e2e/conftest.py
```diff
@@ -40,7 +41,56 @@ def pytest_collection_modifyitems(config, items):
...
 
-# Provide a k8s client to interact with the integration test cluster
-@pytest.fixture(scope='class')
-def k8s_client():
-    return k8s._get_k8s_api_client()
\ No newline at end of file
+
+# Need to use pytest_sessionstart mechanism to store
+# vpc resource information in cache (accessible to worker threads).
+# This implementation is needed to share resources with a "session" scope whether pytest-xdist is used or not.
```
brycahta, line 47: will make this clearer in next revision. Any input would be appreciated!

test/e2e/conftest.py
```diff
@@ -40,7 +41,56 @@ def pytest_collection_modifyitems(config, items):
...
+        resource_name, namespace="default",
+    )
+    k8s.create_custom_resource(ref, resource_data)
+    cr = k8s.wait_resource_consumed_by_controller(ref)
+
+    assert cr is not None
+    assert k8s.get_resource_exists(ref)
+
+    resource = k8s.get_resource(ref)
+    test_resource_values["VPC_ID"] = resource["status"]["vpcID"]
```
brycahta, line 69: removing in next revision

test/e2e/conftest.py
RedbackThomson: I might actually recommend that you simply create the VPC in the `service_bootstrap`, instead. I understand that you would like to test your CRs using the VPC CR, but I think that would make all of the tests depend upon the correctness of your VPC implementation. There is a way to share session-scoped variables across pytest-xdist workers, but it's not fun.

test/e2e/tests/test_route_table.py
```diff
@@ -0,0 +1,134 @@
...
+@pytest.fixture(scope="module")
+def ec2_client():
+    return boto3.client("ec2")
+
+@service_marker
+@pytest.mark.canary
+class TestRouteTable:
+    def get_route_table(self, ec2_client, route_table_id: str) -> dict:
+        try:
+            resp = ec2_client.describe_route_tables()
```
RedbackThomson, line 43:
```suggestion
            resp = ec2_client.describe_route_tables(Filters=[{"Name": "route-table-id", "Values": [route_table_id]}]
```

This should filter by route table ID, and return only a single result - no need for iterating.

test/e2e/tests/test_route_table.py
```diff
@@ -0,0 +1,134 @@
...
+        except Exception as e:
```
RedbackThomson, line 44: Is this try/except necessary? Does `client.describe_route_tables()` throw an exception if there are none in the VPC? I imagine it would return an empty list, instead?

test/e2e/tests/test_route_table.py
```diff
@@ -0,0 +1,134 @@
...
+        for route_table in route_tables:
+            if route_table["RouteTableId"] == route_table_id:
+                return route_table
+
+        return None
+
+    def route_table_exists(self, ec2_client, route_table_id: str) -> bool:
+        return self.get_route_table(ec2_client, route_table_id) is not None
+
+    def test_create_delete(self, ec2_client, pytestconfig):
```
RedbackThomson, line 58: Please see the (`sagemaker-controller` test code)[https://github.com/aws-controllers-k8s/sagemaker-controller/blob/main/test/e2e/tests/test_model.py#L34] for a better way to use fixtures. This code, currently, will not delete the custom resource if an assertion fails. It will assert and then immediately exit the method. Instead, we can use a fixture with a `yield`, which will continue even after the test fails.

test/e2e/tests/test_subnet.py
```diff
@@ -81,12 +53,11 @@ def get_subnet(self, ec2_client, subnet_id: str) -> dict:
     def subnet_exists(self, ec2_client, subnet_id: str) -> bool:
         return self.get_subnet(ec2_client, subnet_id) is not None
 
-    def test_create_delete(self, ec2_client, vpc_resource):
+    def test_create_delete(self, ec2_client, pytestconfig):
```
RedbackThomson, line 56: I know subnet was a previous PR, but please make the same changes as recommended above.

RedbackThomson (commented):
Seems pretty straightforward! No custom logic? Impressive.

Left some comments in your tests.

test/e2e/tests/test_route_table.py
jaypipes: @RedbackThomson is correct:

```
[jaypipes@thelio test-infra]$ aws ec2 describe-route-tables --filters Name=route-table-id,Values=non-existing --output json
{
    "RouteTables": []
}
```

test/e2e/tests/test_route_table.py
```diff
@@ -0,0 +1,134 @@
...
+DELETE_WAIT_AFTER_SECONDS = 10
+
+@pytest.fixture(scope="module")
+def ec2_client():
+    return boto3.client("ec2")
+
+@service_marker
+@pytest.mark.canary
+class TestRouteTable:
+    def get_route_table(self, ec2_client, route_table_id: str) -> dict:
```
jaypipes, line 41: You could more easily make these module-level functions since they don't depend on `self` or the instance/class

test/e2e/tests/test_route_table.py
```diff
@@ -0,0 +1,134 @@
...
+            return None
+
+        route_tables = resp["RouteTables"]
+        for route_table in route_tables:
+            if route_table["RouteTableId"] == route_table_id:
+                return route_table
+
+        return None
+
+    def route_table_exists(self, ec2_client, route_table_id: str) -> bool:
```
jaypipes, line 55: module-level function

test/e2e/tests/test_route_table.py
```diff
@@ -0,0 +1,134 @@
...
+        assert k8s.get_resource_exists(ref)
+
+        expected_msg = "InvalidVpcID.NotFound: The vpc ID 'InvalidVpcId' does not exist"
+        terminal_condition = k8s.get_resource_condition(ref, "ACK.Terminal")
+        # Example condition message:
+        #   InvalidVpcID.NotFound: The vpc ID 'InvalidVpcId' does not exist
+        #   status code: 400, request id: 5801fc80-67cf-465f-8b83-5e02d517d554
+        # This check only verifies the error message; the request hash is irrelevant and therefore can be ignored.
+        assert expected_msg in terminal_condition['message']
+
```
jaypipes, line 134: nit: empty newline

test/e2e/conftest.py
```diff
@@ -40,7 +41,56 @@ def pytest_collection_modifyitems(config, items):
...
+def pytest_sessionfinish(session):
+    worker_input = getattr(session.config, 'workerinput', None)
+    if worker_input is None:
+        # Main thread enters after workers and deletes VPC
+        vpc_resource_name = session.config.cache.get("resource_name", None)
+        ref = k8s.CustomResourceReference(
+            CRD_GROUP, CRD_VERSION, "vpcs",
+            vpc_resource_name, namespace="default",
+        )
+        k8s.delete_custom_resource(ref)
```
jaypipes, line 96: This is actually how I was envisioning a pytest-native plugin that would look for handling resource existence/cleanup. ++

test/e2e/conftest.py
brycahta: > I think that would make all of the tests depend upon the correctness of your VPC implementation.

Right, that would be the intent (as well as other resources added to this cache eventually). If VPC implementation is not correct, then it should be caught during testing. And I also think it's best to test integrating these various pieces together during integration/e2e tests.

> There is a way to share session-scoped variables across pytest-xdist workers, but it's not fun.

Are you referring to writing data to a json file using a FileLock? I linked the latest open ticket related to this in the description above, and I'm fairly certain this implementation is simpler and just as safe.

test/e2e/tests/test_route_table.py
brycahta: Haven't tested yet-- but I'd guess it could throw an exception if credentials expire or aren't available, which is a possible scenario during _local_ testing on a kind cluster

test/e2e/tests/test_route_table.py
brycahta: Good point on the delete, I can deep dive that.

Our use-case for fixtures are not the same-- these are at the module scope not session. I was using vpc fixture in a similar way before these changes, but since more resources will also require vpc the module scope was too narrow

test/e2e/tests/test_route_table.py
brycahta:
```
[7/09/21 9:57:50] ➜  ~ aws ec2 describe-route-tables --profile test-profile

An error occurred (AuthFailure) when calling the DescribeRouteTables operation: AWS was not able to validate the provided access credentials
```

brycahta: Will refine exceptions in next revision

brycahta: In this revision, the exception will also trigger if response has no resources as this is unexpected so I left exception as-is. Let me know if changes are needed/preferred.

A-Hilaly (approved):
Neat :+1:

test/e2e/tests/test_route_table.py
```diff
@@ -0,0 +1,133 @@
...
+
+def get_route_table(ec2_client, route_table_id: str) -> dict:
+    try:
+        resp = ec2_client.describe_route_tables(
+            Filters=[{"Name": "route-table-id", "Values": [route_table_id]}]
+        )
+        route_table = resp["RouteTables"][0]
+    except Exception as e:
+        logging.debug(e)
```
jaypipes, line 45: FYI, this won't log anything particularly useful in the case when the route table does not exist. Instead, you'll get something like `IndexError: 0 out of bounds` because of this line:

```python
route_table = resp["RouteTables"][0]
```

You might want to change that to this:

```python
if len(resp["RouteTables"]) == 0:
     return None
return resp["RouteTables"][0]
```

then your `except` clause will catch valid errors.

test/e2e/tests/test_route_table.py
```diff
@@ -0,0 +1,133 @@
...
+        return None
+
+    return route_table if route_table["RouteTableId"] == route_table_id else None
```
jaypipes, line 48: this isn't necessary when using the filters above.

test/e2e/tests/test_subnet.py
```diff
@@ -33,60 +34,33 @@
...
-        resource_name, namespace="default",
-    )
-    k8s.create_custom_resource(ref, resource_data)
-    cr = k8s.wait_resource_consumed_by_controller(ref)
+def get_subnet(ec2_client, subnet_id: str) -> dict:
+    try:
+        resp = ec2_client.describe_subnets(
+            Filters=[{"Name": "subnet-id", "Values": [subnet_id]}]
+        )
+        subnet = resp["Subnets"][0]
```
jaypipes, line 43: similar comments as above about the IndexError

test/e2e/tests/test_subnet.py
```diff
@@ -33,60 +34,33 @@
...
+    except Exception as e:
+        logging.debug(e)
+        return None
 
-    assert cr is not None
-    assert k8s.get_resource_exists(ref)
+    return subnet if subnet["SubnetId"] == subnet_id else None
```
jaypipes, line 48: not necessary with the use of filters above.

jaypipes (changes requested):

test/e2e/tests/test_route_table.py
brycahta: -_-

test/e2e/tests/test_route_table.py
brycahta: I feel like in this scenario it's "Easier to ask for forgiveness than permission" ([EAFP](https://docs.python.org/3.6/glossary.html#term-eafp)) because:
* we'd usually expect a single result when providing an ID
* we get savings on not needing to do the if check every time, but can still gate the no results scenario
* code is cleaner with less branched logic

The _0 out of bounds_ is not super helpful, but still an extra log to have when debugging tests. In both implementations the lack of a RouteTable would surface a failed `assert` that the resource doesn't exist so I don't think anything is lost there.

I don't think EAFP is more "pythonic" or better than "Look before you leap" ([LBYL](https://docs.python.org/3.6/glossary.html#term-lbyl)), but in this case I think it results in cleaner and clearer code.

Let me know your thoughts and I can update in next revision

jaypipes: > The _0 out of bounds_ is not super helpful, but still an extra log to have when debugging tests. In both implementations the lack of a RouteTable would surface a failed `assert` that the resource doesn't exist so I don't think anything is lost there.

What I'm saying is that "the resource doesn't exist" is not an error and should not log anything.

jaypipes (approved):
:tada:

PR #21, opened 2021-11-16
Add Route to RouteTable using ACK runtime `v0.15.2`

brycahta:
Issue #, if available: https://github.com/aws-controllers-k8s/community/issues/489 

Description of changes:
* Moves `route` from Status to Spec so users can define Routes upon RouteTable creation
* Changes to handle default route created by RouteTable & user-defined routes
* Tests for route


apis/v1alpha1/generator.yaml
```diff
@@ -86,6 +86,16 @@ resources:
       terminal_codes:
         - InvalidVpcID.Malformed
         - InvalidVpcID.NotFound
+        - InvalidParameterValue
+    fields:
+      Routes:
+        custom_field:
+          list_of: Route
```
RedbackThomson, line 93: So clean ❤️

pkg/resource/route_table/hooks.go
```diff
@@ -0,0 +1,354 @@
...
+	svcsdk "github.com/aws/aws-sdk-go/service/ec2"
+)
+
+// RouteAction stores the possible actions that can be performed on
+// any of a Route Table's Routes
+type RouteAction int
+
+const (
+	RouteActionNone RouteAction = iota
+	RouteActionCreate
```
RedbackThomson, line 31: I don't think this is necessary since there is no idea of "updating" a route. This was just an easy enum for the S3 configuration elements, not a critical part of the code path.

pkg/resource/route_table/hooks.go
```diff
@@ -0,0 +1,354 @@
...
+			}
+		}
+	}
+	return action
+}
+
+func compareRoute(
+	a *svcapitypes.Route,
+	b *svcapitypes.Route,
+) *ackcompare.Delta {
```
RedbackThomson, line 114: I think we should generate this method. See the S3 [`sdk_file_end.go.tpl`](https://github.com/aws-controllers-k8s/s3-controller/blob/main/templates/hooks/bucket/sdk_file_end.go.tpl#L103) implementation

pkg/resource/route_table/hooks.go
```diff
@@ -0,0 +1,354 @@
...
+	exit := rlog.Trace("rm.deleteRoute")
+	defer exit(err)
+
+	input := rm.newDeleteRoutePayload(r, c)
+	_, err = rm.sdkapi.DeleteRouteWithContext(ctx, input)
+	rm.metrics.RecordAPICall("DELETE", "DeleteRoute", err)
+	return err
+}
+
+func (rm *resourceManager) newCreateRoutePayload(
```
RedbackThomson, line 255: +1 generate dynamically again

pkg/resource/route_table/hooks.go
```diff
@@ -0,0 +1,354 @@
...
+		input.SetTransitGatewayId(*c.TransitGatewayID)
+	}
+	if c.VPCPeeringConnectionID != nil {
+		input.SetVpcPeeringConnectionId(*c.VPCPeeringConnectionID)
+	}
+
+	return input
+}
+
+func (rm *resourceManager) newDeleteRoutePayload(
```
RedbackThomson, line 303: +1 generate dynamically again

pkg/resource/route_table/sdk.go
```diff
@@ -414,6 +414,17 @@ func (rm *resourceManager) sdkCreate(
 	}
 
 	rm.setStatusDefaults(ko)
+	if rm.requiredFieldsMissingForCreateRoute(&resource{ko}) {
+		return nil, ackerr.NotFound
+	}
+
+	if len(desired.ko.Spec.Routes) > 0 {
+		//desired routes are overwritten by RouteTable's default route
+		ko.Spec.Routes = append(ko.Spec.Routes, desired.ko.Spec.Routes...)
```
RedbackThomson, line 423: This is essentially late initialization?

RedbackThomson (commented):
Awesome use of the custom list fields - glad to see it came in handy. Left some inline comments, especially regarding removing generated code that you would otherwise need to maintain.

pkg/resource/route_table/hooks.go
brycahta: yup will clean this up

pkg/resource/route_table/hooks.go
brycahta: Saw that and had it originally, but have a couple reservations:
* Template file is more difficult to develop and maintain relative to a native go file
* Keeping this code in `hook.go` makes it clearer and more explicit what logic is being injected by service controller teams

With that said, can you help me see what kinda benefit there are to moving this to a `.tpl`?

brycahta:
Taking a look at failing `Subnet` test. Might be related to me adding a public subnet to the SharedVPC

pkg/resource/route_table/sdk.go
brycahta: Correct! Unfortunately, couldn't use Late Init because of some limitations such as:

[Unlike Map, updating individual element of a list is not supported. LateInitializing complete list is supported.](https://github.com/aws-controllers-k8s/code-generator/blob/main/pkg/generate/code/late_initialize.go#L77)

With RouteTable, I only need to late initialize a single element (the default route) not the entire list

pkg/resource/route_table/hooks.go
```diff
@@ -0,0 +1,354 @@
...
+	}
+	if c.NetworkInterfaceID != nil {
+		input.SetNetworkInterfaceId(*c.NetworkInterfaceID)
+	}
+	if c.TransitGatewayID != nil {
+		input.SetTransitGatewayId(*c.TransitGatewayID)
+	}
+	if c.VPCPeeringConnectionID != nil {
+		input.SetVpcPeeringConnectionId(*c.VPCPeeringConnectionID)
+	}
```
jaypipes, line 298: Did you want to override `input.DryRun = false` always?

test/e2e/tests/test_route_table.py
```diff
@@ -54,9 +76,13 @@ def test_create_delete(self, ec2_client):
         resource_name = random_suffix_name("route-table-test", 24)
         test_vpc = get_bootstrap_resources().SharedTestVPC
         vpc_id = test_vpc.vpc_id
+        igw_id = test_vpc.public_subnets.route_table.internet_gateway.internet_gateway_id
+        test_cidr_block = "100.68.0.0/18"
```
bwagner5, line 80: any reason to use this CIDR versus an RFC1918, private CIDR like 10.0.0.0/24

bwagner5 (commented):
nice! A few comments and questions, feel free to take them or leave them

pkg/resource/route_table/hooks.go
```diff
@@ -0,0 +1,354 @@
...
+			if !desiredRoute {
+				if err = rm.deleteRoute(ctx, latest, *l); err != nil {
+					return err
+				}
+			}
+
+		}
+	}
+
+	return nil
```
jaypipes, line 85: I want the desired routes to be an idempotent "replacement" of Routes on a RouteTable. In other words, I want to support an update code path for Routes.

I'd prefer to see this code first **gather** the routes that need to be added and deleted. You can do this like so:

```go
toAdd := []*svcapitypes.Route{}
toDelete := []*svcapitypes.Route{}

for _, desiredRoute := range desired.ko.Spec.Routes {
	if latestRoute := matchingRoute(desiredRoute); latestRoute != nil {
		if routeChanged(desiredRoute, latestRoute) {
			// "update" the route by deleting the old route and adding
			// the new route
			toDelete = append(toUpdate, latestRoute)
			toAdd = append(toAdd, desiredRoute)
		}
	} else {
		// It's a brand new route...
		toAdd = append(toAdd, desiredRoute)
	}
}
for _, latestRoute := range latest.ko.Spec.Routes {
	if desiredRoute := matchingRoute(latestRoute); desiredRoute == nil {
		toDelete = append(toDelete, latestRoute)
	}
}
```

Now, once you've gathered the routes that need to be added and deleted, you can do the following in a controlled manner:

```go
for _, route := range toDelete {
	rlog.Debug("deleting route from route table", ...)
	deleteRoute(ctx, r, route)
}
for _, route := range toAdd {
	rlog.Debug("adding route to route table", ...)
	createRoute(ctx, r, route)
}
```

I'll leave it up to you to implement the `matchingRoute` and `routeChanged` functions referenced above but would like to point out the following quote from the CreateRoute API documentation:

> You must specify one of the following targets: internet gateway or virtual private gateway, NAT instance, NAT gateway, VPC peering connection, network interface, egress-only internet gateway, or transit gateway. 

The above means that you can likely make a number of short-circuits in your `matchingRoute` and `routeChanged` function implementations because you know that, for example, if `CarrierGatewayID` is not nil, you know that you don't need to look for differences in, say, `NatGatewayID` because those fields are mutually exclusive.

pkg/resource/route_table/hooks.go
```diff
@@ -0,0 +1,354 @@
...
+	if ackcompare.HasNilDifference(a.LocalGatewayID, b.LocalGatewayID) {
+		delta.Add("Route.LocalGatewayID", a.LocalGatewayID, b.LocalGatewayID)
+	} else if a.LocalGatewayID != nil && b.LocalGatewayID != nil {
+		if *a.LocalGatewayID != *b.LocalGatewayID {
+			delta.Add("Route.LocalGatewayID", a.LocalGatewayID, b.LocalGatewayID)
+		}
+	}
+	if ackcompare.HasNilDifference(a.NatGatewayID, b.NatGatewayID) {
+		delta.Add("Route.NatGatewayID", a.NatGatewayID, b.NatGatewayID)
+	} else if a.NatGatewayID != nil && b.NatGatewayID != nil {
```
jaypipes, line 181: to fix in a later PR... we should capitalize `NAT` in `NatGatewayID`

test/e2e/tests/test_route_table.py
brycahta: haha using this one because it was the first one I came across in [create-route documentation](https://awscli.amazonaws.com/v2/documentation/api/latest/reference/ec2/create-route.html):

```
--destination-cidr-block (string)

The IPv4 CIDR address block used for the destination match. Routing decisions are based on the most specific match. We modify the specified CIDR block to its canonical form; for example, if you specify 100.68.0.18/18 , we modify it to 100.68.0.0/18 .
```

Would a private CIDR be clearer?

pkg/resource/route_table/hooks.go
```diff
@@ -0,0 +1,354 @@
...
+}
+
+func (rm *resourceManager) syncRoutes(
+	ctx context.Context,
+	desired *resource,
+	latest *resource,
+) (err error) {
+	rlog := ackrtlog.FromContext(ctx)
+	exit := rlog.Trace("rm.syncRoutes")
+	defer exit(err)
```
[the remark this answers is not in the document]
jaypipes: @bwagner5 @brycahta `exit` is a convention used for tracing throughout the runtime. For example:

https://github.com/aws-controllers-k8s/runtime/blob/5afdd00a4bf2587efa53dbf7693f4a550f43fea4/pkg/runtime/reconciler.go#L203

I'd prefer we keep it as-is considering all the other controllers and the runtime use this convention.

pkg/resource/route_table/hooks.go
brycahta: Should be fixed once I auto-generate per ACK-Nick's comment

test/e2e/tests/test_route_table.py
bwagner5: I think so :)

pkg/resource/route_table/hooks.go
brycahta: I'm not sure if I can take advantage of the dynamic generation here. I am populating a `CreateRouteInput` using a `svcapitypes.Route`; however, not all of the fields in Route are valid in `CreateRouteInput`. For example, Route.Origin and Route.State are set server-side(so not part of the create request), but the following invalid lines would generate:

```
if c.Origin != nil {
		input.SetState(*c.Origin)
	}
if c.State != nil {
		input.SetState(*c.State)
	}
```

Although these fields cannot be decided by the user, they are still important to surface in the spec which is why I used a slice of `Route` instead of slice of `CreateRouteInput` (briefly mentioned in https://github.com/aws-controllers-k8s/community/issues/489 when first discussed)

RedbackThomson: > Although these fields cannot be decided by the user, they are still important to surface in the spec

Why is that? It seems that if they cannot be set they should be a status field?

Rather than using `list_of: Route`, I might recommend a `list_of: CreateRouteInput`?

brycahta: > Why is that? It seems that if they cannot be set they should be a status field?

To clarify my wording, I'm of the mind these fields should be surfaced to the user, but my only option is via the Spec. I agree since they cannot be set, then they should be a status field; however, I don't see a clean way to separate these (i.e. some fields in Route Spec set by user and some set by server-side).

I read up on the issue to deal with server-side returned values https://github.com/aws-controllers-k8s/community/issues/812 to see how to handle it in my situation, and the direction here seems to be okay with directly populating desired.Spec with server-returned vals-- please correct me if I'm wrong.

I could use `list_of: CreateRouteInput` , but then `Origin` and `State` won't be exposed to the user. Origin might be less important, but I thinking seeing route's state as active or not is important. The trade off here seems to be have proper CX (Spec fields should be user-editable) vs. expose potentially helpful info about Route.. are you leaning towards the former?

RedbackThomson: > I could use `list_of: CreateRouteInput` , but then `Origin` and `State` won't be exposed to the user. Origin might be less important, but I thinking seeing route's state as active or not is important. The trade off here seems to be have proper CX (Spec fields should be user-editable) vs. expose potentially helpful info about Route.. are you leaning towards the former?

Yeah I understand your concern. Thinking about the longer-term picture, we cannot go through the one way door of adding `State` to the spec if we will never consider it to be editable. I would prefer it is left out altogether than to add possibly misleading fields. It is possible that in the future we can also have `State` information in a relevant part of the status - I'd rather work toward that.

brycahta: I'm not opposed to switching to `list_of: CreateRouteInput` ; however, the underlying issue of _something in the spec that is not editable_ will still be present due to RouteTable's default route:

```
  Spec:
  Routes:
    Destination CIDR Block:  172.31.0.0/16 --> Can maybe change?
    Gateway ID:              local --> CANNOT CHANGE
    Origin:                  CreateRouteTable
    State:                   active
```

When users try to change it, a terminal condition occurs (see newly added smoke test) telling them it's not possible. If we are okay with that CX, then could we do something similar regarding `Origin` and `State` fields for all Routes? Last note, is users also cannot create a new route with Gateway ID: local despite it being a valid value here.. not sure if that's any concern or not.

RedbackThomson: > RouteTable's default route

Yes this is a limitation imposed on us as a late initialized field, but for all the other routes every spec field is configurable. I'm ignoring the default route because it's the outlier and not much we can control with that. Terminal conditions are gross because they're validated at runtime. If we can protect users from editing fields that will have no effect (`State`), or will otherwise definitely cause a terminal condition, then I'll be happy.

templates/hooks/route_table/sdk_file_end.go.tpl
```diff
@@ -0,0 +1,26 @@
+{{ $CRD := .CRD }}
```
brycahta, line 1: `.tpl` only brings suffering. I would rather hard-code the funcs as before & 'watch' aws-sdk-go repo on these shapes for changes daily than maintain this.

pkg/resource/route_table/hooks.go
brycahta: In that case, I think it makes sense to just go with `list_of: CreateRouteInput` for now. I'll get some comments on these latest commits and make the updates after.


Brainstorming ideas for later improvement:
* improve Late Init to handle single element in list and *some* fields in a resource
* resource relationship with Route eventually
* is it possible to use `list_of: CreateRouteInput` in the Spec and keep `list_of:Routes` in the Status (it's already in RouteTable.Status by default) and apply overrides to ignore non-status fields? 
  * One "issue" that comes to mind is there's no Id for a Route, but any other attributes that could be used to identify are editable in Spec (CIDR block, Gateway ID)

templates/hooks/route_table/sdk_file_end.go.tpl
jaypipes: Trust us, you will prefer the generated/template approach over the long term :) Yes, `.tpl` files suck to work with, agreed, but any time you can remove a piece of manually-maintained code is a win in this game.

jaypipes (approved):
All good with me. If @RedbackThomson and @bwagner5 want to send it on its merry way, that would be cool. :)

pkg/resource/route_table/hooks.go
brycahta: https://github.com/aws-controllers-k8s/code-generator/pull/245

templates/hooks/route_table/sdk_file_end.go.tpl
brycahta: Yeah, I can definitely see the benefits and don't mind the `.tpl` in _code-generator_ mostly because all the refs are contained in the file. It's just frustrating to edit this file when you need to constantly refer to the file in _code-generator_ that eventually consumes it to understand context. With that said, @jaypipes @RedbackThomson  any helpful extensions or general feedback that has helped you guys working with templates would be greatly appreciated! I'm using Go Template Support in VSCode right now, but really only helps with syntax highlighting.


A side note: if I move the Create/Delete Payload funcs here (per Nick's comment), there would still need to be something from the model hard-coded in this template (whether it's the op name or index). In which case, an update to the API _could_ still break this even if it was templated. Although it'd be a smaller risk of breakage than hard-coding the funcs as I'm doing now, I think the trade-off (clearer code) is worth it for now. Either way-- probably low risk that this API changes any time soon (unless...?)

pkg/resource/route_table/hooks.go
jaypipes: Hmmm, after looking into this further, I don't actually think Route should be an element of RouteTable.Spec. Instead, I think Route should be a full-fledged CRD itself and the Origin and State fields should go in the Route.Status, with all the other fields going in Route.Spec. In other words, let's treat Route just like any other custom resource.

Sorry I had not realized this before today. :(

jaypipes:
```yaml
resources:
  RouteTable:
    RouteStatuses:
      from:
        operation: DescribeRouteTableOutput.Routes
      is_read_only: true
```

something like the above might work to get a readonly copy of the Route information (including State/Origin) into the RouteTable.Status.RouteStatuses field?

brycahta: Since Routes is already part of RouteTable.Status by default so I think this would just need to be a rename. Let me try that and re-generate

templates/hooks/route_table/sdk_file_end.go.tpl
RedbackThomson: > With that said, @jaypipes @RedbackThomson any helpful extensions or general feedback that has helped you guys working with templates would be greatly appreciated! I'm using Go Template Support in VSCode right now, but really only helps with syntax highlighting.

I found that banging my head against the wall for hours trying to get it working really helped

pkg/resource/route_table/hooks.go
brycahta: Here's what `kubectl describe routetables` looks like with the latest changes:

```
Spec:
  Routes:
    Destination CIDR Block:  172.31.0.0/16
    Gateway ID:              local
    Destination CIDR Block:  100.68.0.0/16
    Gateway ID:              igw-c31af0a4
  Vpc ID:                    vpc-e5867283
Status:
  Ack Resource Metadata:
    Owner Account ID:  515336597380
  Conditions:
  Owner ID:  515336597380
  Route Statuses:
    Destination CIDR Block:  172.31.0.0/16
    Gateway ID:              local
    Origin:                  CreateRouteTable
    State:                   active
    Destination CIDR Block:  100.68.0.0/16
    Gateway ID:              igw-c31af0a4
    Origin:                  CreateRoute
    State:                   active
  Route Table ID:            rtb-0c23ffe9b126e794f
Events:                      <none>
```

generator.yaml
```diff
@@ -86,6 +87,25 @@ resources:
...
+        - InvalidParameterValue
+    fields:
+      RouteStatuses:
+        from:
+          operation: DescribeRouteTables
+          path: RouteTables.Routes
+        is_read_only: true
+      Routes:
+        custom_field:
+          list_of: CreateRouteInput
```
jaypipes, line 99: Don't hesitate to put comments into the generator.yaml file indicating _why_ certain configurations were set on the resource. Remember to inform your future self about decisions made in the past :)

jaypipes (approved):
Looks awesome, thanks @brycahta! @RedbackThomson feel free to send this on its merry way if you agree.

pkg/resource/route_table/hooks.go
```diff
@@ -0,0 +1,228 @@
...
+	c svcapitypes.CreateRouteInput,
+) (err error) {
+	rlog := ackrtlog.FromContext(ctx)
+	exit := rlog.Trace("rm.createRoute")
+	defer exit(err)
+
+	input := rm.newCreateRouteInput(c)
+	// Routes should only be configurable for the
+	// RouteTable in which they are defined
+	input.RouteTableId = r.ko.Status.RouteTableID
```
RedbackThomson, line 156: +1

RedbackThomson (approved):
Awesome. This is in a great spot, now! I really don't see any compromises for users or for yourself as the maintainer.

Ship it!

/lgtm

PR #39, opened 2022-03-14
Add `ElasticIPAddress` CRD

RedbackThomson:
Issue #, if available: https://github.com/aws-controllers-k8s/community/issues/1218

Description of changes:
Creates the `ElasticIPAddress` custom resource, from `AllocateAddress` and `DeallocateAddress`. This API uses either `AllocationID` or `PublicIP` as the identifier depending on whether the domain is set to `"vpc"` (for EC2-VPC) or otherwise (for EC2-Classic), respectively. For the adoption, I have chosen to set `AllocationID` as the primary key for now, since EC2-VPC is the recommended method for managing resources in EC2.


RedbackThomson:
@brycahta What do you think of the name? Should this be renamed to `ElasticIPAddress` or even just `Address`? I used the term `ElasticIP` since the console shows these resources as `Elastic IP Addresses`.

config/crd/bases/ec2.services.k8s.aws_elasticips.yaml
```diff
@@ -0,0 +1,177 @@
...
+kind: CustomResourceDefinition
+metadata:
+  annotations:
+    controller-gen.kubebuilder.io/version: v0.7.0
+  creationTimestamp: null
+  name: elasticips.ec2.services.k8s.aws
+spec:
+  group: ec2.services.k8s.aws
+  names:
+    kind: ElasticIP
```
A-Hilaly, line 13: This is a duplicate file of `config/crd/bases/ec2.services.k8s.aws_elasticipaddresses.yaml`

A-Hilaly (commented):
Looks good! Left one comment below

apis/v1alpha1/elastic_ip_address.go
```diff
@@ -0,0 +1,108 @@
...
+	//
+	// You cannot use a network border group with EC2 Classic. If you attempt this
+	// operation on EC2 Classic, you receive an InvalidParameterCombination error.
+	NetworkBorderGroup *string `json:"networkBorderGroup,omitempty"`
+	// The ID of an address pool that you own. Use this parameter to let Amazon
+	// EC2 select an address from the address pool. To specify a specific address
+	// from the address pool, use the Address parameter instead.
+	PublicIPv4Pool *string `json:"publicIPv4Pool,omitempty"`
+	// The tags to assign to the Elastic IP address.
+	TagSpecifications []*TagSpecification `json:"tagSpecifications,omitempty"`
```
jaypipes, line 54: Can we rename this just `Tags`?

apis/v1alpha1/elastic_ip_address.go
```diff
@@ -0,0 +1,108 @@
...
+	// The ID of a customer-owned address pool. Use this parameter to let Amazon
+	// EC2 select an address from the address pool. Alternatively, specify a specific
+	// address from the address pool.
+	CustomerOwnedIPv4Pool *string `json:"customerOwnedIPv4Pool,omitempty"`
+	// Indicates whether the Elastic IP address is for use with instances in a VPC
+	// or instances in EC2-Classic.
+	//
+	// Default: If the Region supports EC2-Classic, the default is standard. Otherwise,
+	// the default is vpc.
+	Domain *string `json:"domain,omitempty"`
```
jaypipes, line 37: Recommend removing this field (since we should not support EC2-Classic) and always override it to `vpc`

jaypipes (changes requested):

generator.yaml
```diff
@@ -85,6 +86,23 @@ ignore:
...
 operations:
+  AllocateAddress:
+    operation_type: 
+    - Create
+    resource_name: ElasticIPAddress
+  DescribeAddresses:
+    operation_type:
+    - List
+    resource_name: ElasticIPAddress
+    custom_check_required_fields_missing_method: customCheckRequiredFieldsMissing
```
brycahta, line 97: Is it necessary to use `custom_check_required_fields_missing_method` here vs. resource having a pre_build_request hook?

generator.yaml
```diff
@@ -127,6 +145,52 @@ resources:
...
+      AllocationId:
+        is_primary_key: true
+        print:
+          name: ALLOCATION-ID
+      PublicIp:
+        print:
+          name: PUBLIC-IP
+    list_operation:
+      match_fields:
+      # PublicIP and AllocationID are two ways of identifying the same resource
```
brycahta, line 179: Per the comment on line 19, we won't be supporting ec2-classic because it will be deprecated soon™️

helm/crds/ec2.services.k8s.aws_elasticips.yaml
```diff
@@ -0,0 +1,177 @@
+
```
brycahta, line 1: leftover file from the rename?

brycahta (changes requested):

apis/v1alpha1/elastic_ip_address.go
RedbackThomson: All the other EC2 resources have this as `TagSpecifications` as well. It's not a slice of key-value pairs like you would expect. The format is instead:
```yaml
spec:
  tagSpecifications:
  - resourceType: "elastic-ip"
     tags:
     - key: Name
       value: "My-Elastic-IP"   
```

Replace the `resourceType` field with `"vpc"`, `"nat-gateway"`, etc. So they're at least all consistently inconsistent with the standard practice. I think these should all be replaced with a `Tags` field and the `resourceType` inject by the controller before making a request.

generator.yaml
RedbackThomson: It just exits a little bit earlier, and simplifies the logic in that code block. But no, I could reduce it down to one.

brycahta: Gotcha, in that case could you move it to a hook? That would align with existing code and long-term plan to move away from operation-centric configs (I know some are necessary atm).

brycahta (approved):
looks solid to me

generator.yaml
```diff
@@ -131,6 +146,47 @@ resources:
...
+      - UnsupportedProtocol
+      - ValidationError
+    fields:
+      AllocationId:
+        is_primary_key: true
+        print:
+          name: ALLOCATION-ID
+      PublicIp:
+        print:
+          name: PUBLIC-IP
```
jaypipes, line 177: We wanted to remove this, right?

RedbackThomson: Is it not still something that's relevant,even for EC2-VPC? I thought the point of an Elastic IP address was that you were reserving this Public IP within the AWS IP space. I just don't use it as the identifier anywhere.

brycahta (approved):
/lgtm

PR #41, opened 2022-03-18
Refactor e2e tests to use assertion helper class 

brycahta:
Issue #, if available: N/A

Description of changes:
* Adds a helper class, `Ec2Validator`, to assert resource data/state against aws apis more easily
* This refactor will also make implementing resource reference e2e test easier (follow-up PR)


test/e2e/tests/helper.py
```diff
@@ -0,0 +1,103 @@
...
+#
+# or in the "license" file accompanying this file. This file is distributed
+# on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+# express or implied. See the License for the specific language governing
+# permissions and limitations under the License.
+
+"""Helper functions for ec2 tests
+"""
+
+class Ec2Validator:
```
RedbackThomson, line 17: Please use the proper casing for an initialism
```suggestion
class EC2Validator:
```

test/e2e/tests/test_dhcp_options.py
```diff
@@ -84,14 +67,15 @@ def test_create_delete(self, ec2_client):
 
         time.sleep(CREATE_WAIT_AFTER_SECONDS)
 
-        # Check DHCP Options exists
-        assert dhcp_options_exist(ec2_client, resource_id)
+        # Check DHCP Options exists in AWS
+        ec2_validator = Ec2Validator(ec2_client)
+        ec2_validator.assert_dhcp_options(resource_id)
```
RedbackThomson, line 72: While I think checking for existence is useful, I believe we can harden our tests by checking their spec matches. Especially for code with custom hooks. Perhaps, we should have methods in the `EC2Validator` that get the resources, and then we can just check `assert dhcp_options is not None` (and all of our spec assertions as well).

test/e2e/tests/helper.py
```diff
@@ -0,0 +1,103 @@
...
+    def __init__(self, ec2_client):
+        self.ec2_client = ec2_client
+
+    def assert_dhcp_options(self, dhcp_options_id: str, exists=True):
```
RedbackThomson, line 21: This is asserting the existence of DHCP Options
```suggestion
    def assert_dhcp_options_exists(self, dhcp_options_id: str, exists=True):
```

test/e2e/tests/test_route_table.py
```diff
@@ -189,26 +152,21 @@ def test_crud_route(self, ec2_client):
...
+        ec2_validator.assert_route(resource_id, igw_id, "CreateRoute")
         
         # Update Route
+        default_cidr = "10.0.0.0/16"
         updated_cidr = "192.168.1.0/24"
         patch = {"spec": {"routes": [
                     {
                         #Default route cannot be changed
-                        "destinationCIDRBlock": default_cidr,
+                        "destinationCIDRBlock": "10.0.0.0/16",
```
RedbackThomson, line 169: Should this be using `default_cidr`?

RedbackThomson (changes requested):

test/e2e/tests/helper.py
brycahta: Can you clarify? I purposefully omitted "exists" in the func name after looking at APIGateway tests so I could use the same func for both assertions

test/e2e/tests/test_dhcp_options.py
brycahta: I agree with raising the quality of these tests, but should be addressed in a separate PR/issue. Do you know of an issue (or parent issue) to improve e2e testing or should I open one for this controller?

Once this is merged, my plan is to open a follow-up PR with a resource-reference e2e test modeled after ApiGatewayV2 (checks existence of referenced resources) so we can get some coverage on the feature. Then, I can work on improving e2e tests to check resources at a more granular level across the entire suite. Lmk if you have any other context (i.e. if some automated test-gen in the works, you prefer to finish vpc enhancements before e2e test improvements, etc.) or if you think it can be approached better.

test/e2e/tests/test_route_table.py
brycahta: good catch; yup!

brycahta:
> ec2-kind-e2e — Job failed.

TransitGateway test failed because the resource still exists, but in the 'deleting' state. I checked the account and it has been deleted. This also happened to me when I tested locally, but was hoping it was transient. Not sure why it's occurring now, so will deep dive and improve the test in next revision.


#### Logs
```
tests/test_transit_gateway.py:77: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

self = <e2e.tests.helper.Ec2Validator object at 0x7f57cafb6070>
tgw_id = 'tgw-0bbc70e488bf84a9a', exists = False

    def assert_transit_gateway(self, tgw_id: str, exists=True):
        res_found = False
        try:
            aws_res = self.ec2_client.describe_transit_gateways(TransitGatewayIds=[tgw_id])
            res_found = len(aws_res["TransitGateways"]) > 0
        except self.ec2_client.exceptions.ClientError:
            pass
>       assert res_found is exists
E       AssertionError

tests/helper.py:85: AssertionError
=========================== short test summary info ============================
FAILED tests/test_transit_gateway.py::TestTGW::test_create_delete - Assertion...
=================== 1 failed, 13 passed in 120.92s (0:02:00) ===================
```

test/e2e/tests/test_dhcp_options.py
RedbackThomson: > Do you know of an issue (or parent issue) to improve e2e testing or should I open one for this controller?

This is on a per-controller basis. So if there isn't one for EC2, you can create one. No automated test-gen in the works atm :(

test/e2e/tests/helper.py
RedbackThomson: An assertion checks one thing against another. "Assert DHCP Options" doesn't say what it's asserting against, so it's an incomplete statement. "Assert DHCP Options Exists" tells you what it's actually asserting, so that's why it's preferable. It should be updated in the APIGateway tests too, if that's where you got it :)

brycahta: If it sounds incomplete, then I can explicitly pass in the `exists` parameter when checking for existence:
```
# check existence
assert_dhcp_options(id, exists=True)

# check doesn't exist
assert_dhcp_options(id, exists=False)
```

[`APIGateway`](https://github.com/aws-controllers-k8s/apigatewayv2-controller/blob/main/test/e2e/tests/helper.py#L119) has 2 separate funcs for this, `assert_res__is_present`/`asset_res_is_deleted` (maybe this should be renamed for accuracy).

IMO 1 func is enough and clearer, but if we're trying to align styles across controllers to make maintaining easier, then I'm all for it. Details should go in an issue.

RedbackThomson (approved):
Only merging this because you promised me better tests in the near future haha. Counting on it!

/lgtm

PR #48, opened 2022-03-30
VPC attachment for `InternetGateway`

RedbackThomson:
Closes https://github.com/aws-controllers-k8s/community/issues/1216

Description of changes:
Add support for attaching an `InternetGateway` to a `VPC`. The EC2 API exposes a list of `Attachments` when you describe `InternetGateway`, but manual testing shows that you can only attach a gateway to a single VPC - so this is treated as a list of at most one element. I had to create a new VPC for integration tests, since the default VPC we use for all of the other tests already contains an IGW, and you can have at most 1 per VPC.


vijtrip2 (approved):
Looks good.

test/e2e/tests/test_internet_gateway.py
```diff
@@ -70,5 +72,75 @@ def test_create_delete(self, ec2_client):
...
+        k8s.patch_custom_resource(ref, updates)
+        time.sleep(MODIFY_WAIT_AFTER_SECONDS)
+
+        # Check there are no attachments on Internet Gateway
+        igw = ec2_validator.get_internet_gateway(resource_id)
+
+        # In the case where it shows the attachment as being in detached state
+        if len(igw["Attachments"]) == 1:
+            assert igw["Attachments"][0]["VpcId"] == vpc_id
+            assert igw["Attachments"][0]["State"] == "detached"
```
brycahta, line 134: Should this be an `or` check?  "detached" or "detaching"

test/e2e/tests/test_internet_gateway.py
```diff
@@ -70,5 +72,75 @@ def test_create_delete(self, ec2_client):
 
         time.sleep(DELETE_WAIT_AFTER_SECONDS)
 
+        # Check Internet Gateway no longer exists in AWS
+        ec2_validator.assert_internet_gateway(resource_id, exists=False)
+
+    def test_vpc_association(self, ec2_client):
+        resource_name = random_suffix_name("ig-ack-test", 24)
+
+        test_vpc = get_bootstrap_resources().EmptyVPC
```
brycahta, line 81: If this VPC is only used here, then would you be opposed to creating the VPC in this test and exercising the resource reference code paths?

brycahta (approved):
changes look solid 👍

just a couple nits/questions on the test

test/e2e/tests/test_internet_gateway.py
RedbackThomson: It only takes a few fractions of a second to detach. Our sleep above should suffice for this.

test/e2e/tests/test_internet_gateway.py
RedbackThomson: Yeah happy to move it into the test as a fixture, itself. Not sure about using it as a ref, though. I think we need to have a holistic discussion around testing references in general. Adding a test for every field, and then another test for every ref, is going to blow up the size of our tests and I don't know how much value it actually brings (since the reference resolution code is generated deterministically). I'll leave that for another day, though.

brycahta (approved):
/lgtm

PR #55, opened 2022-05-25
Generate `Instance` CRD

brycahta:
Issue #, if available: https://github.com/aws-controllers-k8s/community/issues/1196

Description of changes:
* Instance Controller create/delete functionality with smoke test


brycahta:
I will be handling update/modify code as well as exceptions and TerminalConditions in the next PR. Until then, Instance is NOT able to achieve ResourceSynced state because of "rolling" late initialization for fields such as `BlockDeviceMappings`. 

**Detailed Details**
* If a user does not specify `BlockDeviceMappings` in the Spec, then neither the Create Response nor the subsequent Describe Response contains a populated `BlockDeviceMappings` field (good!)
* After multiple Describe calls (usually by the 3rd call here:  https://github.com/aws-controllers-k8s/runtime/blob/main/pkg/runtime/reconciler.go#L237 because LateInit is enabled), the `BlockDeviceMappings` field is eventually updated with a server-side value, ex:

```
//prev
BlockDeviceMappings: []

//updated
BlockDeviceMappings: [{
              DeviceName: "/dev/xvda",
              Ebs: {
                AttachTime: 2022-05-25 17:20:03 +0000 UTC,
                DeleteOnTermination: true,
                Status: "attaching",
                VolumeId: "vol-008cf94ebf53b06ec"
              }
            }],
```
* This triggers an update (not yet impl) and thus resource does not reach Sync state.
* Note, I have tried LateInitializing `BlockDeviceMappings`, but per the first point it does not apply here.

apis/v1alpha1/instance.go
```diff
@@ -0,0 +1,382 @@
...
+	LicenseSpecifications []*LicenseConfigurationRequest `json:"licenseSpecifications,omitempty"`
+	// The maximum number of instances to launch. If you specify more instances
+	// than Amazon EC2 can launch in the target Availability Zone, Amazon EC2 launches
+	// the largest possible number of instances above MinCount.
+	//
+	// Constraints: Between 1 and the maximum number you're allowed for the specified
+	// instance type. For more information about the default limits, and how to
+	// request an increase, see How many instances can I run in Amazon EC2 (http://aws.amazon.com/ec2/faqs/#How_many_instances_can_I_run_in_Amazon_EC2)
+	// in the Amazon EC2 FAQ.
+	MaxCount *int64 `json:"maxCount,omitempty"`
```
RedbackThomson, line 155: These should be removed from the spec?

apis/v1alpha1/instance.go
```diff
@@ -0,0 +1,382 @@
...
+	State *InstanceState `json:"state,omitempty"`
+	// The reason for the most recent state transition.
+	// +kubebuilder:validation:Optional
+	StateReason *StateReason `json:"stateReason,omitempty"`
+	// The reason for the most recent state transition. This might be an empty string.
+	// +kubebuilder:validation:Optional
+	StateTransitionReason *string `json:"stateTransitionReason,omitempty"`
+	// Any tags assigned to the instance.
+	// +kubebuilder:validation:Optional
+	Tags []*Tag `json:"tags,omitempty"`
```
RedbackThomson, line 345: This is a duplicate of `TagSpecifications` essentially?

test/e2e/tests/test_instance.py
```diff
@@ -0,0 +1,139 @@
...
+            if 'Description' in image:
+                if INSTANCE_AMI in image['Description']:
+                    return image['ImageId']
+    except Exception as e:
+        logging.debug(e)
+
+@service_marker
+@pytest.mark.canary
+class TestInstance:
+    def test_create_delete(self, ec2_client):
```
RedbackThomson, line 91: Creating the resource, asserting, and deleting all within the test method itself inherently leaves it open to dangling resources. If an assertion fails, the delete methods will never be called, and the instances will be left running in the account. Please refer to the [EKS Cluster tests](https://github.com/aws-controllers-k8s/eks-controller/blob/main/test/e2e/tests/test_cluster.py#L57-L88) to see how I use fixtures to get around this. Essentially by creating and deleting within a fixture, the assertions can be made independently and not impact them.

RedbackThomson (changes requested):
I just want to make sure we're not duplicating fields between Spec and Status. I saw `VPCID` defined in Status, but I don't see it defined in Spec - which makes me wonder how I would set this value in the first place.

Also left a comment inline about testing methodology.

test/e2e/tests/test_instance.py
brycahta: Good call; addressing in next revision

brycahta:
>  I saw VPCID defined in Status, but I don't see it defined in Spec - which makes me wonder how I would set this value in the first place.

Good catch! By default, `RunInstances` API will use the **default VPC** and a random subnet. If you want to select which VPC+Subnet to use, then the *subnetID* needs to be specified in the request. Basically there's no way to pass a VPCID in the request. https://docs.aws.amazon.com/cli/latest/reference/ec2/run-instances.html

apis/v1alpha1/instance.go
brycahta: After some deep dive, it looks like the code is **NOT** expected to remove these from Spec. These configs are at the operational level and don't seem to impact or be coupled with CRD Spec/Status in any way.  Another data point: the `override_values` in `rds-controller` are not present in the Spec because they never were. These fields apply to the operation only.

I know we have efforts to [move away from operation-centric configs](https://github.com/aws-controllers-k8s/community/issues/1185), which would be one way to address this. Or do you want me to open a new issue?

apis/v1alpha1/instance.go
RedbackThomson: > This should probably be reworked so Instance Controller can only tag Instance; I'm thinking remove TagSpecifications altogether and expose Tags in Spec. What do you think?

Yes! This should be the standard for all of the EC2 APIs, IMO

pkg/resource/instance/hooks.go
```diff
@@ -0,0 +1,77 @@
...
+}
+
+// updateTagSpecificationsInCreateRequest updates the
+// TagSpecification in RunInstancesInput (Create) to
+// remove Tags unrelated to 'instance' and merge
+// defined in the resource's Spec to the Create Request (RunInstancesInput).
+// This is needed for a couple reasons:
+//	- Expose 'Tags' in Spec for a clearer CX
+//  - Instance Controller should not be able to edit other resources
+func updateTagSpecificationsInCreateRequest(r *resource,
```
brycahta, line 42: @RedbackThomson this hook code is to support both `TagSpecifications` and `Tags` in the Spec, and I think this is necessary given the config limitations of _code-generator_. Open to other config suggestions if it's possible to support only Tags today; lmk.

**Detailed Details**
In an ideal world, I'd like to **ignore** `RunInstancesInput.TagSpecifications`; however, I need to reference this field in order to add `Tags` to the Spec. The add-to-Spec logic is coupled with *input shapes* and the only instance-related operation that had `Tags` shape in its input shape was `RunInstancesInput`

apis/v1alpha1/instance.go
brycahta: Agreed on this standard. 

However, not sure how straightforward or possible it is with existing _code-generator_ functionality. I left a comment with some more details, but tldr is I don't think I can ignore TagSpecifications and add Tags to Spec simultaneously. Might be worth investing in a better solution since this will apply to most EC2 APIs

pkg/resource/instance/hooks.go
```diff
@@ -0,0 +1,55 @@
...
+		for _, desiredTag := range r.ko.Spec.Tags {
+			// Add in tags defined in the Spec
+			tag := &svcsdk.Tag{}
+			if desiredTag.Key != nil && desiredTag.Value != nil {
+				tag.SetKey(*desiredTag.Key)
+				tag.SetValue(*desiredTag.Value)
+			}
+			instanceTags = append(instanceTags, tag)
+		}
+		desiredTagSpecs.SetResourceType("instance")
```
RedbackThomson, line 51: nit: Move into a constant

RedbackThomson (approved):
Great job!

vijtrip2 (approved):

PR #58, opened 2022-07-12
Internet Gateway terminal codes

LikithaVemulapalli:

Description of changes:

Made the following changes in order to map exceptions for **internet gateway** resource in ec2-controller.
      - Added terminal codes in the "generator.yaml" for internet gateway resource.
      - Added few integration tests for internet gateway resource that cover invalid and malformed vpc-id.

### DIFF
--- a/generator.yaml
+++ b/generator.yaml
@@ -235,11 +235,7 @@ resources:
   InternetGateway:
     exceptions:
       terminal_codes:
-      - InvalidInternetGatewayID.NotFound
-      - InvalidInternetGatewayId.Malformed
       - InvalidVpcId.Malformed
-      - InvalidVpcID.NotFound
-      - InvalidParameterValue
     fields:
       Tags:
         from:
--- a/generator.yaml
+++ b/generator.yaml
@@ -237,8 +237,6 @@ resources:
       terminal_codes:
       - InvalidInternetGatewayID.NotFound
       - InvalidInternetGatewayId.Malformed
-      - Resource.AlreadyAssociated
-      - Gateway.NotAttached
       - InvalidVpcId.Malformed
       - InvalidVpcID.NotFound
       - InvalidParameterValue
--- a/generator.yaml
+++ b/generator.yaml
@@ -233,6 +233,15 @@ resources:
       sdk_read_many_post_build_request:
         template_path: hooks/elastic_ip_address/sdk_read_many_post_build_request.go.tpl
   InternetGateway:
+    exceptions:
+      terminal_codes:
+      - InvalidInternetGatewayID.NotFound
+      - InvalidInternetGatewayId.Malformed
+      - Resource.AlreadyAssociated
+      - Gateway.NotAttached
+      - InvalidVpcId.Malformed
+      - InvalidVpcID.NotFound
+      - InvalidParameterValue
     fields:
       Tags:
         from:
--- a/test/e2e/tests/test_internet_gateway.py
+++ b/test/e2e/tests/test_internet_gateway.py
@@ -183,40 +183,9 @@ class TestInternetGateway:
         # Check Internet Gateway no longer exists in AWS
         ec2_validator.assert_internet_gateway(resource_id, exists=False)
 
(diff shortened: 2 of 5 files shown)


brycahta (commented):
/ok-to-test

generator.yaml
```diff
@@ -233,6 +233,15 @@ resources:
       sdk_read_many_post_build_request:
         template_path: hooks/elastic_ip_address/sdk_read_many_post_build_request.go.tpl
   InternetGateway:
+    exceptions:
+      terminal_codes:
+      - InvalidInternetGatewayID.NotFound
+      - InvalidInternetGatewayId.Malformed
+      - Resource.AlreadyAssociated
```
brycahta, line 240: I don't think exceptions such as `Resource.AlreadyAssociated` and `Gateway.NotAttached` are possible given the [attach/detach implementation](https://github.com/aws-controllers-k8s/ec2-controller/blob/main/pkg/resource/internet_gateway/hook.go#L39). Am I missing context or can we remove?

test/e2e/tests/test_internet_gateway.py
```diff
@@ -181,4 +181,67 @@ def test_vpc_association(self, ec2_client, empty_vpc):
         time.sleep(DELETE_WAIT_AFTER_SECONDS)
 
         # Check Internet Gateway no longer exists in AWS
-        ec2_validator.assert_internet_gateway(resource_id, exists=False)
\ No newline at end of file
+        ec2_validator.assert_internet_gateway(resource_id, exists=False)
+
+    def test_terminal_condition_invalid_vpcid(self):
+        test_resource_values = REPLACEMENT_VALUES.copy()
+        resource_name = random_suffix_name("ig-invalid-vpcid", 24)
```
brycahta, line 188: `resource_name` is being assigned a `vpcid` value in both tests. Can you update so the name reflects the *internet gateway* resource?

brycahta (changes requested):
Nice first PR! 🎉 

Had a couple comments on minor things, but overall very clean

generator.yaml
LikithaVemulapalli: Yes I agree with the statement that these exceptions are not possible considering our implementation. We cannot test them, thought that we can include them in terminal codes, I can remove them!

test/e2e/tests/test_internet_gateway.py
LikithaVemulapalli: I added resource_name values as per the failure scenarios, will update with the names that reflects the internet gateway resource.

generator.yaml
brycahta: I'm thinking we just include the exceptions that are possible/testable:
```
      - InvalidVpcId.Malformed
      - InvalidVpcID.NotFound
```

@RedbackThomson @jaypipes  do y'all have other thoughts/opinions?

vijtrip2: My 2 cents:

TL;DR: Vet carefully the exceptions being marked as terminal because missing an exception as terminal is better than marking a recoverable exception as terminal.

---
 
Terminal exceptions cause the reconciliation to stop. The exceptions added to the terminal exception should be vetted carefully. 
In case of some terminal exceptions not being added to this block, controller will just retry with exponential backoff. However if some recoverable exception gets added here, the reconciliation will cease.

---

Also NotFound errors should not be marked as terminal because we have seen customer submit their manifests out of order and some resources being created after their dependent resource.
In resource reference we handle this scenario, but have seen customer complaints in the past(they weren't using references)

brycahta: Appreciate the insights and feedback @vijtrip2 -- very helpful 👍 

All your points make sense to me. Basically, we only want `InvalidVpcId.Malformed` as a terminal exception for `InternetGateway` (Likitha & I have deep dived the other exceptions already) because reconciliation cannot "correct" this to a valid state. We can always add more exceptions to be terminal later should any issues arise. Feel free to correct anything/lmk if that sounds good. Thanks again

brycahta (commented):
/lgtm

PR #65, opened 2022-08-01
Add Rules to Security Group

brycahta:
Issue #, if available: https://github.com/aws-controllers-k8s/community/issues/1326

Description of changes:
* Adds rules to SecurityGroup Spec and Status with tests


brycahta:
> kubectl describe securitygroups

```
Name:         securitygroup-ack-test
Namespace:    default
Labels:       <none>
Annotations:  <none>
API Version:  ec2.services.k8s.aws/v1alpha1
Kind:         SecurityGroup
Metadata:
  Creation Timestamp:  2022-07-27T16:09:51Z
  Finalizers:
    finalizers.ec2.services.k8s.aws/SecurityGroup
  Generation:  1
  Managed Fields:
    API Version:  ec2.services.k8s.aws/v1alpha1
    Fields Type:  FieldsV1
    fieldsV1:
      f:metadata:
        f:annotations:
          .:
          f:kubectl.kubernetes.io/last-applied-configuration:
      f:spec:
        .:
        f:description:
        f:egressRules:
        f:ingressRules:
        f:name:
        f:vpcID:
    Manager:      kubectl-client-side-apply
    Operation:    Update
    Time:         2022-07-27T16:09:51Z
    API Version:  ec2.services.k8s.aws/v1alpha1
    Fields Type:  FieldsV1
    fieldsV1:
      f:metadata:
        f:finalizers:
          .:
          v:"finalizers.ec2.services.k8s.aws/SecurityGroup":
      f:spec:
        f:tags:
      f:status:
        .:
        f:ackResourceMetadata:
          .:
          f:ownerAccountID:
          f:region:
        f:conditions:
        f:id:
        f:rules:
    Manager:         main
    Operation:       Update
    Time:            2022-07-27T16:09:54Z
  Resource Version:  87996072
  Self Link:         /apis/ec2.services.k8s.aws/v1alpha1/namespaces/default/securitygroups/securitygroup-ack-test
  UID:               03cbb789-e1fa-4e59-930a-a53fb2556621
Spec:
  Description:  My security Group
  Egress Rules:
    From Port:    80
    Ip Protocol:  tcp
    Ip Ranges:
      Cidr IP:      172.31.0.0/16
      Description:  test EGRESS sg rules
    To Port:        80
  Ingress Rules:
    From Port:    80
    Ip Protocol:  tcp
    Ip Ranges:
      Cidr IP:      172.31.0.0/16
      Description:  test INGRESS sg rules
    To Port:        80
  Name:             MyFirstGroup
  Vpc ID:           vpc-0101f5ec94e71f602
Status:
  Ack Resource Metadata:
    Owner Account ID:  515336597380
    Region:            us-west-2
  Conditions:
    Last Transition Time:  2022-07-27T16:09:54Z
    Message:               Resource synced successfully
    Reason:
    Status:                True
    Type:                  ACK.ResourceSynced
  Id:                      sg-0c8c63bd3f66af8f5
  Rules:
    cidrIPv4:                172.31.0.0/16
    Description:             test EGRESS sg rules
    From Port:               80
    Ip Protocol:             tcp
    Is Egress:               true
    Security Group Rule ID:  sgr-0715febb47a689187
    To Port:                 80
    cidrIPv4:                172.31.0.0/16
    Description:             test INGRESS sg rules
    From Port:               80
    Ip Protocol:             tcp
    Is Egress:               false
    Security Group Rule ID:  sgr-0c155c0eeebba694a
    To Port:                 80
Events:                      <none>
```

apis/v1alpha1/generator.yaml
```diff
@@ -351,6 +354,17 @@ resources:
       # support EC2-VPC only
       Id:
         is_primary_key: true
+      IngressRules:
+        custom_field:
+          list_of: IpPermission
+      EgressRules:
+        custom_field:
+          list_of: IpPermission
```
RedbackThomson, line 362: Hmmm. So on the console these are referred to as "Inbound" and "Outbound" rules. But in the CloudFormation and CDK resources these are called "Ingress" and "Egress". +1 for "Ingress/Egress"

pkg/resource/security_group/hook.go
```diff
@@ -14,9 +14,150 @@
...
+	ackrtlog "github.com/aws-controllers-k8s/runtime/pkg/runtime/log"
 	svcsdk "github.com/aws/aws-sdk-go/service/ec2"
 )
 
+func (rm *resourceManager) addRulesToSpec(
+	ko *svcapitypes.SecurityGroup,
+	resp *svcsdk.SecurityGroup,
+) {
+	ko.Spec.IngressRules = nil
+	ko.Spec.EgressRules = nil
```
RedbackThomson, line 31: Why are you nil-ing them out, here?

templates/hooks/security_group/sdk_create_post_set_output.go.tpl
```diff
@@ -0,0 +1,11 @@
+	
+	if rm.requiredFieldsMissingForSGRule(&resource{ko}) {
+		return nil, ackerr.NotFound
+	}
+	if err = rm.removeDefaultEgressRule(ctx, &resource{ko}); err != nil {
+		return nil, err
+	}
```
RedbackThomson, line 7: I think I disagree with removing the default egress rule. If a user does not explicitly specify the rules, we should stick to the server-side default. (I believe) The vast majority of security groups will have the default egress rule, and so this is an expected behaviour by users. That being said, if they specify any egress rules, then we should override and delete the default.

pkg/resource/security_group/hook.go
```diff
@@ -14,9 +14,150 @@
...
+		}
+		ko.Spec.EgressRules = specEgress
+	}
+}
+
+func (rm *resourceManager) addRulesToStatus(
+	ko *svcapitypes.SecurityGroup,
+	ctx context.Context,
+) (err error) {
+	ko.Status.Rules = nil
```
RedbackThomson, line 53: nit: Not necessary, since we override on line 86

pkg/resource/security_group/hook.go
```diff
@@ -14,9 +14,150 @@
...
+			}
+		}
+	}
+
+	if err = rm.createSecurityGroupRules(ctx, desired, toAddIngress, toAddEgress); err != nil {
+		return err
+	}
+	if err = rm.deleteSecurityGroupRules(ctx, latest, toDeleteIngress, toDeleteEgress); err != nil {
+		return err
+	}
```
RedbackThomson, line 156: My logic is that we should delete before we create for two reasons:
1. Updating a rule requires that it be removed before the updated version be added
2. If there is an error with adding new rules, it fails after it has already deleted the old ones. This is safer, since you have disabled the older rules.

test/e2e/tests/test_security_group.py
```diff
@@ -111,4 +111,128 @@ def test_terminal_condition(self):
...
+        test_resource_values["VPC_ID"] = vpc_id
+        test_resource_values["SECURITY_GROUP_DESCRIPTION"] = "TestSecurityGroupRule-create-delete"
+        
+        # Create Security Group CR with ingress and egress rules
+        test_resource_values["IP_PROTOCOL"] = "tcp"
+        test_resource_values["FROM_PORT"] = "80"
+        test_resource_values["TO_PORT"] = "80"
+        test_resource_values["CIDR_IP"] = "172.31.0.0/16"
+        test_resource_values["DESCRIPTION_INGRESS"] = "test ingress rule"
+        test_resource_values["DESCRIPTION_EGRESS"] = "test egress rule"
```
RedbackThomson, line 132: If these are hard-coded, then please just hardcode them in the YAML file

test/e2e/tests/test_security_group.py
```diff
@@ -111,4 +111,128 @@ def test_terminal_condition(self):
...
+                                "description": "test egress update"
+                            }
+                        ]
+                    }
+        ]}}
+        _ = k8s.patch_custom_resource(ref, patch)
+        time.sleep(CREATE_WAIT_AFTER_SECONDS)
+
+        # assert patched state
+        print(f'{ref}')
```
RedbackThomson, line 196: nit: Remove print statements (they aren't shown in test logs anyway)

RedbackThomson (changes requested):
Lots of custom code, but all looks great. Left some comments inline

pkg/resource/security_group/hook.go
brycahta: Hmm not sure. I think it was leftover from testing; let me try to remove

test/e2e/tests/test_security_group.py
brycahta: Would you mind if I kept the assignments? I think it's clearer to see what is being created within the test and it doesn't require knowledge about how the tests work.

pkg/resource/security_group/hook.go
brycahta: Well said and makes sense to me! Going to add a comment with the same

pkg/resource/security_group/hook.go
jaypipes: @brycahta what if there are existing non-nil values for `ko.Spec.IngressRules` and `ko.Spec.EgressRules`? Do we want to nil those out even if `resp.IpPermissions` or `resp.IpPermissionsEgress` are nil?

pkg/resource/security_group/hook.go
```diff
@@ -14,9 +14,150 @@
...
+	"context"
+
+	svcapitypes "github.com/aws-controllers-k8s/ec2-controller/apis/v1alpha1"
+
+	ackcompare "github.com/aws-controllers-k8s/runtime/pkg/compare"
+	ackrtlog "github.com/aws-controllers-k8s/runtime/pkg/runtime/log"
 	svcsdk "github.com/aws/aws-sdk-go/service/ec2"
 )
 
+func (rm *resourceManager) addRulesToSpec(
```
jaypipes, line 26: would be great to have docstrings for these methods... :)

pkg/resource/security_group/hook.go
jaypipes: we would not set `ko.Status.Rules` to nil if there was an error on line 83, though. And I think it's more appropriate to leave `ko.Status.Rules` as-is in the case of an error in `DescribeSecurityGroupRules`, no?

pkg/resource/security_group/hook.go
```diff
@@ -39,3 +180,146 @@ func updateTagSpecificationsInCreateRequest(r *resource,
...
+		latest, err = rm.sdkFind(ctx, latest)
+		if err != nil {
+			return nil, err
+		}
+	}
+
+	return latest, nil
+}
+
+// contains returns true if security group rule
```
jaypipes, line 295: perhaps `containsRule` might be a bit more descriptive of this function's purpose?

templates/hooks/security_group/sdk_create_post_set_output.go.tpl
jaypipes: Agree with @RedbackThomson here.

pkg/resource/security_group/hook.go
```diff
@@ -39,3 +180,146 @@ func updateTagSpecificationsInCreateRequest(r *resource,
...
+		if len(delta.Differences) == 0 {
+			return true
+		}
+	}
+	return false
+}
+
+func toBoolPtr(boolean bool) *bool {
+	return &boolean
+}
```
jaypipes, line 317: unused AFAICT?

test/e2e/tests/test_security_group.py
jaypipes: @brycahta did you want to remove these?

brycahta: Yes, I was "resolving" as I fixed things locally. Will be removed in next revision

pkg/resource/security_group/hook.go
brycahta: I don't think we want to nil those out; I removed setting to nil in both `addRulesToStatus` and `addRulesToSpec`.

jaypipes (commented):
This is looking very good, @brycahta thank you! I left just one question inline for you. Depending on answer for that, I'd be ready to merge this.

pkg/resource/security_group/delta.go
```diff
@@ -48,6 +48,12 @@ func newResourceDelta(
 			delta.Add("Spec.Description", a.ko.Spec.Description, b.ko.Spec.Description)
 		}
 	}
+	if !reflect.DeepEqual(a.ko.Spec.EgressRules, b.ko.Spec.EgressRules) {
+		delta.Add("Spec.EgressRules", a.ko.Spec.EgressRules, b.ko.Spec.EgressRules)
+	}
+	if !reflect.DeepEqual(a.ko.Spec.IngressRules, b.ko.Spec.IngressRules) {
+		delta.Add("Spec.IngressRules", a.ko.Spec.IngressRules, b.ko.Spec.IngressRules)
+	}
```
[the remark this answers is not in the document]
brycahta: Thanks for the callout. I did some local testing:
* rearranging the order of `egressRules` and `ingressRules` did not trigger a delta difference
* rearranging the order of the rules within `ingressRules` did trigger a delta difference

brycahta:
> This is looking very good, @brycahta thank you! I left just one question inline for you. Depending on answer for that, I'd be ready to merge this.

@jaypipes Found a couple issues while testing locally hence the `do-not-merge` label. It's making me think `LateInitialize` is not the correct approach for `EgressRules`. Working on the fixes now; can discuss details on Slack once the outage is over, if you're interested.

pkg/resource/security_group/delta.go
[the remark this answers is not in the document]
brycahta: Discussed offline; neither scenario should trigger a delta difference because there is no change in **data**. Will create issue to add hook for `delta preCompare` (or similar) to sort prior to checking for equality. This needs to be done for any/all resources with list of struct fields

pkg/resource/security_group/hook.go
brycahta: I was wrong for `addRulesToSpec`. This is needed because `sdkUpdate` will make a call to `sdkFind` to get the latest resource state which then executes this code. If the latest resource state DOES NOT contain rules, then the rules from the previous resource state will persist as they are never nil-ed out.

pkg/resource/security_group/hook.go
```diff
@@ -39,3 +178,195 @@ func updateTagSpecificationsInCreateRequest(r *resource,
...
+	rm.setStatusDefaults(ko)
+
+	if delta.DifferentAt("Spec.IngressRules") || delta.DifferentAt("Spec.EgressRules") {
+		if err := rm.syncSGRules(ctx, desired, latest); err != nil {
+			return nil, err
+		}
+		latest, err = rm.sdkFind(ctx, latest)
+		if err != nil {
+			return nil, err
+		}
```
jaypipes, line 328: Is this additional call to `sdkFind()` necessary? Won't `latest` already have the latest `Spec` information?

jaypipes (commented):
@brycahta one final question inline...

pkg/resource/security_group/hook.go
brycahta: Yes, it's necessary. For example, `latest` may contain an egress rule while `desired` does not. After call to `syncSGRules`, that egress rule should be removed server-side but the `latest` Spec will still contain it. The call to `sdkFind()` then refreshes both Spec/Status rules with latest data from server, resulting in `latest` having its Spec and Status updated to no longer include the deleted rule.

jaypipes: Why not just return `desired` here then?

brycahta: If I returned desired, then I don't think the `Status.Rules` will be an accurate representation of the rules existing server-side. That data isn't available until after `syncSGRules()` call and then `sdkFind()` updates both Spec+Status with the latest data.

jaypipes: Hmm, I see... instead of calling the full `sdkFind` method, then, what about just setting `Status.Rules` at the end of `syncSGRules` by calling `DescribeSecurityGroupRules` then?

brycahta: I need the full `sdkFind` method (i.e. update both Spec and Status) on `Update` because of the scenario when user updates SecurityGroup with their own EgressRule. 

The intended behavior is remove default egress rule if user provides one and since this is done automatically, the `desired` Spec will still include the default egress rule if we only update `Status.Rules` in `Update`

jaypipes (approved):
This is great, @brycahta thank you! I would love to see a tutorial (or addition to the github.com/aws-controllers-k8s/examples` repository) that shows how to use this new functionality, since it's functionality that isn't covered directly by the ACK resource reference.

PR #68, opened 2022-08-08
Fix Internet Gateway terminal codes

LikithaVemulapalli:

Description of changes: 

- Removed `InvalidVpcId.Malformed` error from terminal codes for **Internet Gateway** resource.
- Removed Integration tests for the `InvalidVpcId.Malformed` terminal code.

Initially, I thought `InvalidVpcId.Malformed` should be included in the terminal codes. Later, observed that Internet Gateway resource being created after adding the integration tests for the malformed vpc id terminal code. This is expected as per our code because initially the resource will be created with no vpc attachment and later if we pass VPC ID, it will be attached to the resource. After discussions and local testing I observed that if you add malformed VPC ID it will create a resource and if you try to update the VPC ID to a valid one, it will create a new resource and leave the initial one in dangling state that cannot be deleted, so we decided not to include malformed VPC ID  for Internet Gateway resource.

### DIFF
--- a/apis/v1alpha1/ack-generate-metadata.yaml
+++ b/apis/v1alpha1/ack-generate-metadata.yaml
@@ -1,5 +1,5 @@
 ack_generate_info:
-  build_date: "2022-08-02T21:22:57Z"
+  build_date: "2022-08-08T17:04:17Z"
   build_hash: 87477ae8ca8ac6ddb8c565bbd910cc7e30f55ed0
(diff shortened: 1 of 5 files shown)


brycahta (commented):
/ok-to-test

brycahta (approved):
/lgtm

Thanks for the fix and clear explanation 👍

PR #69, opened 2022-08-09
Terminal code mappings for RouteTable

LikithaVemulapalli:

Description of changes:

- Removed `InvalidVpcId.NotFound`, `InvalidVpcId.Malformed` and `InvalidParameterValue` terminal codes for **Route Table** resource.
- Removed Integration test that exists for `InvalidVpcId.NotFound` terminal code.

I tried adding integration test for the `InvalidParameterValue` by passing an invalid parameter, but it leaves a dangling **VPC** resource which is incorrect, tested locally by using patch and found dangling resource being created. Considering these scenarios, removed the above mentioned terminal codes for **Route Table** resource.


brycahta (commented):
/ok-to-test

brycahta:
@LikithaVemulapalli couple questions:

* Is `InvalidVpcId.Malformed` removed because if Vpc is malformed the `InvalidVpcId.NotFound` will be returned instead?
* Which parameter(s) were under test for `InvalidParameterValue`?

> I tried adding integration test for the InvalidParameterValue by passing an invalid parameter, but it leaves a dangling VPC resource which is incorrect, tested locally by using patch and found dangling resource being created. Considering these scenarios, removed the above mentioned terminal codes for Route Table resource.

I'm a little confused on the dangling resource. VPC should already exist when creating RouteTable (unless you're testing with references). Aside from that, after looking thru the code, I think `InvalidParameterValue` should still map to a terminal condition. I would expect providing invalid Routes to result in no resource (RouteTable) being created and an error returned: https://github.com/aws-controllers-k8s/ec2-controller/blob/main/pkg/resource/route_table/sdk.go#L423

LikithaVemulapalli:
> @LikithaVemulapalli couple questions:
> 
> * Is `InvalidVpcId.Malformed` removed because if Vpc is malformed the `InvalidVpcId.NotFound` will be returned instead?

Yes, if we provide a malformed VPC ID, we get `InvalidVpcId.NotFound` error, so removed both the terminal codes.

> * Which parameter(s) were under test for `InvalidParameterValue`?

I used `destinationCIDRBlock` to check the `InvalidParameterValue` terminal code.

> > I tried adding integration test for the InvalidParameterValue by passing an invalid parameter, but it leaves a dangling VPC resource which is incorrect, tested locally by using patch and found dangling resource being created. Considering these scenarios, removed the above mentioned terminal codes for Route Table resource.
> 
> I'm a little confused on the dangling resource. VPC should already exist when creating RouteTable (unless you're testing with references). Aside from that, after looking thru the code, I think `InvalidParameterValue` should still map to a terminal condition. I would expect providing invalid Routes to result in no resource (RouteTable) being created and an error returned: https://github.com/aws-controllers-k8s/ec2-controller/blob/main/pkg/resource/route_table/sdk.go#L423

I passed an invalid CIDR block both initially and cross-checked by updating the patch routes with invalid CIDR block, I thought it's leaving a dangling resource (VPC) after running route table test file as VPC is being created and I'm getting cleanup errors as VPC cannot be deleted due to dependencies.

brycahta (approved):
/lgtm

PR #73, opened 2022-08-24
refactor vpc_endpoint test to use fixture

brycahta:
Issue #, if available: N/A

Description of changes: Refactoring e2e tests to use `pytest.fixture` for the **resource under test**. Otherwise, assertion failures within the test may not lead to successful resource cleanup.

### DIFF
--- a/test/e2e/tests/test_vpc_endpoint.py
+++ b/test/e2e/tests/test_vpc_endpoint.py
@@ -34,39 +34,51 @@ ENDPOINT_SERVICE_NAME = "com.amazonaws.%s.s3" % REGION
 CREATE_WAIT_AFTER_SECONDS = 10
 DELETE_WAIT_AFTER_SECONDS = 10
 
+@pytest.fixture
+def simple_vpc_endpoint():
+    test_resource_values = REPLACEMENT_VALUES.copy()
+    resource_name = random_suffix_name("vpc-endpoint-test", 24)
+    test_vpc = get_bootstrap_resources().SharedTestVPC
+    vpc_id = test_vpc.vpc_id
+
+    test_resource_values["VPC_ENDPOINT_NAME"] = resource_name
+    test_resource_values["SERVICE_NAME"] = ENDPOINT_SERVICE_NAME
+    test_resource_values["VPC_ID"] = vpc_id
+
+    # Load VPC Endpoint CR
+    resource_data = load_ec2_resource(
+        "vpc_endpoint",
+        additional_replacements=test_resource_values,
+    )
+    logging.debug(resource_data)
+
+    # Create k8s resource
+    ref = k8s.CustomResourceReference(
+        CRD_GROUP, CRD_VERSION, RESOURCE_PLURAL,
+        resource_name, namespace="default",
+    )
+    k8s.create_custom_resource(ref, resource_data)
+    time.sleep(CREATE_WAIT_AFTER_SECONDS)
+
+    cr = k8s.wait_resource_consumed_by_controller(ref)
+    assert cr is not None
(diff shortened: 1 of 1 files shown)


A-Hilaly (approved):
👍 
/lgtm

test/e2e/tests/test_vpc_endpoint.py
```diff
@@ -34,39 +34,51 @@
...
+    assert k8s.get_resource_exists(ref)
+
+    yield (ref, cr)
+
+    # Try to delete, if doesn't already exist
+    try:
+        _, deleted = k8s.delete_custom_resource(ref, 3, 10)
+        assert deleted
+    except:
+        pass
```
jaypipes, line 74: @brycahta @A-Hilaly this isn't actually what we should be doing in the post-`yield` block. We actually should be checking for existence using the CLI and then deleting the resource using the Python CLI, not the Kubernetes CLI, since we're calling `k8s.delete_custom_resource()` already below on line 90 (original line 78). This post-`-yield` is a safety-guard and we should use the Python CLI to clean up resources that potentially were not properly cleaned up in the normal test case flow.

So, this really should be:

```python
if ec2_validator.vpc_exists(ref.vpc_id):
    try:
        ec2_client.delete_vpc(vpc_id)
    except:
        pass
```
```

A-Hilaly: The deletion check exists in [L96 using boto3](https://github.com/aws-controllers-k8s/ec2-controller/pull/73/files#diff-c6ba7f72232c14efea0a035fbc2c16bb21aa3e4ca7e928000b502c3fab2a9abeR89-R96)

That's why we don't panic here if the `assert deleted` raises a exception.

brycahta: I thought this post-`yield` safety guard is for when we prematurely fail a test prior to calling `k8s.delete_custom_resource()`. This way we guarantee it is always called and we exercise/test the resource's delete code path as a bonus. 

> since we're calling k8s.delete_custom_resource() already below on line 90 (original line 78). This post--yield is a safety-guard and we should use the Python CLI to clean up resources that potentially were not properly cleaned up in the normal test case flow

This sounds like you want a guard in case `k8s.delete_custom_resource()` does not work properly-- am I misreading? If so, that's a bug within the resource's delete logic and should be fixed not rely on fixture.

jaypipes: @brycahta yes, that is correct. The guard here should be in case the Kubernetes delete custom resource fails unexpectedly.

A-Hilaly: I guess i miss understood how this should work. Sadly we use this pattern everywhere..

A-Hilaly: @jaypipes i believe that, in case tests fail, it makes more sense to call `k8s_client.delete` rather than `boto3.client.delete`. For these reasons:
- Controller might call create again if a reconcilliation is still happening for a reason...
- Most the the controller delete call don't have any issues. We only need to make sure that the CR is deleted.
- Most of the dangling resources we see in e2e tests are because of assertions breaking the tests and not allowing `k8s_client.delete`

brycahta: @jaypipes I see. Correct me if I'm wrong, but it sounds like we need the safeguard for 2 reasons:
* in case we fail an assert and exit a test prematurely prior to hitting delete resource code
* in case `k8s.delete_custom_resource()` fails unexpectedly

Tying back to your suggestion, we should use Python CLI/Boto3 to delete so we can handle both of these cases with the same code. Is this accurate?

PR #75, opened 2022-08-24
Refactor `DHCPOptions` tests to use fixture

A-Hilaly:
Refactoring e2e tests to use `pytest.fixture` for the resource under
test. Prior to these changes, assertion failures could leave dangling
resources in the testing account.

Signed-off-by: Amine Hilaly <hilalyamine@gmail.com>

### DIFF
--- a/test/e2e/tests/test_dhcp_options.py
+++ b/test/e2e/tests/test_dhcp_options.py
@@ -30,40 +30,62 @@ DEFAULT_WAIT_AFTER_SECONDS = 5
 CREATE_WAIT_AFTER_SECONDS = 10
 DELETE_WAIT_AFTER_SECONDS = 10
 
+@pytest.fixture
+def simple_dhcp_options(request):
+    replacements = REPLACEMENT_VALUES.copy()
+    resource_name = random_suffix_name("dhcp-opts-test", 24)
+
+    replacements["DHCP_OPTIONS_NAME"] = resource_name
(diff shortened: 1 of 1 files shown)


test/e2e/tests/test_dhcp_options.py
```diff
@@ -30,40 +30,62 @@
...
+    resource_name = random_suffix_name("dhcp-opts-test", 24)
+
+    replacements["DHCP_OPTIONS_NAME"] = resource_name
+    replacements["DHCP_KEY_1"] = "domain-name"
+    replacements["DHCP_VAL_1"] = "ack-example.com"
+    replacements["DHCP_KEY_2"] = "domain-name-servers"
+    replacements["DHCP_VAL_2_1"] = "10.2.5.1"
+    replacements["DHCP_VAL_2_2"] = "10.2.5.2"
+
+    marker = request.node.get_closest_marker("dhcp_options_data")
```
jaypipes, line 45: this works just fine, of course, but you could name the marker `fixture_overrides` or similarly-generic term and have all these fixture marker overrides use the same terminology instead of having each marker named differently.

jaypipes (approved):

test/e2e/tests/test_dhcp_options.py
A-Hilaly: Nice suggestion. I will submit next PRs using `fixture_overrides` and send another one to rename the one that already exists in `main`. /cc @brycahta

PR #77, opened 2022-08-25
Refactor `InternetGateway` tests to use fixture

A-Hilaly:
Refactoring e2e tests to use pytest.fixture for the resource under
test. Prior to these changes, assertion failures could leave dangling
resources in the testing account.

Signed-off-by: Amine Hilaly <hilalyamine@gmail.com>

### DIFF
--- a/test/e2e/conftest.py
+++ b/test/e2e/conftest.py
@@ -13,11 +13,20 @@
 
 import boto3
 import pytest
+import time
+import logging
+
 from acktest.aws.identity import get_region
+from acktest.resources import random_suffix_name
+from acktest.k8s import resource as k8s
+from e2e import CRD_GROUP, CRD_VERSION, load_ec2_resource
+from e2e.replacement_values import REPLACEMENT_VALUES
+
+VPC_CREATE_WAIT_AFTER_SECONDS = 10
+VPC_RESOURCE_PLURAL = "vpcs"
 
 def pytest_addoption(parser):
     parser.addoption("--runslow", action="store_true", default=False, help="run slow tests")
-
 
 def pytest_configure(config):
     config.addinivalue_line(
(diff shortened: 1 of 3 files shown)


test/e2e/tests/test_internet_gateway.py
```diff
@@ -25,7 +25,7 @@
 from e2e.tests.helper import EC2Validator
 from e2e.bootstrap_resources import get_bootstrap_resources
 
-from .test_vpc import RESOURCE_PLURAL as VPC_RESOURCE_PLURAL, CREATE_WAIT_AFTER_SECONDS as VPC_CREATE_WAIT
+from .test_vpc import simple_vpc
```
brycahta, line 28: Should this be moved to `conftest` so we don't have cross-test dependency?

A-Hilaly: Good idea ++

test/e2e/conftest.py
```diff
@@ -45,4 +54,51 @@ def pytest_collection_modifyitems(config, items):
 @pytest.fixture(scope="module")
 def ec2_client():
     region = get_region()
-    return boto3.client("ec2", region)
\ No newline at end of file
+    return boto3.client("ec2", region)
+
+@pytest.fixture(scope="module")
```
brycahta, line 59: Checked the logs and I believe setting scope to module is the root cause of the test failure. 

Because this is being invoked once per *module* as opposed to once per *function*, `TestVpc` module will reuse this vpc for all its tests. However, we want a new vpc to be created *per test*,  Ex: `test_enable_attributes `  creates a new vpc with attributes set to true.

brycahta: On the bright side, this failure resulted in **zero** dangling resources 🎉

A-Hilaly: Gotcha! Thank you for investigating this. Removed the module scope

brycahta (approved):
/lgtm

PR #81, opened 2022-08-26
Update tags support for Subnet

LikithaVemulapalli:
Issue #, if available: https://github.com/aws-controllers-k8s/community/issues/1369

Description of changes:

-  Added custom code that supports updating tags for **Subnet** resource.
- Added integration test to check whether tags are being updated as expected.


test/e2e/tests/test_subnet.py
```diff
@@ -97,6 +97,8 @@ def test_create_delete(self, ec2_client):
         test_resource_values["VPC_ID"] = vpc_id
         # CIDR needs to be within SharedTestVPC range and not overlap other subnets
         test_resource_values["CIDR_BLOCK"] = "10.0.255.0/24"
+        test_resource_values["KEY"] = "createKey"
```
brycahta, line 100: related to comment above; I'm assuming if you don't define these then the tests will fail? If so, can you add an assert check on this tag? If we define the data, then we should validate it

test/e2e/tests/test_subnet.py
```diff
@@ -134,6 +139,84 @@ def test_create_delete(self, ec2_client):
         # Check Subnet no longer exists in AWS
         ec2_validator.assert_subnet(resource_id, exists=False)
 
+    def test_update_tags(self, ec2_client):
```
brycahta, line 142: Update tags test looks solid 👍  

Would you consider expanding the test coverage on tags a bit and updating this to a `test_crud_tags`? After your update check:
1.  **add** a 2nd tag to the resource --> validate data
2. **delete** a tag from the resource --> validate data

This will exercise the paths of adding a tag post-creation and also deleting a tag while resource is active

brycahta (changes requested):
Nice work @LikithaVemulapalli ! Very excited to get this tagging functionality in.

pkg/resource/subnet/hooks.go
```diff
@@ -243,6 +250,131 @@ func toStrPtr(str string) *string {
...
+		if err != nil {
+			return err
+		}
+	}
+
+	return nil
+}
+
+// compareTags compares the tags and adds the difference to delta
+func compareTags(
```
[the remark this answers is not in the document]
LikithaVemulapalli: I haven't tried using go templates, but I like this idea as the code will be consistent with the other compare functions, will make changes accordingly. Thank you!

test/e2e/tests/test_subnet.py
LikithaVemulapalli: As mentioned earlier, tests won't fail if tags are not defined, I added here so that they align with the resource file. But I can delete tags from `test_create_delete` as I'm planning to add tests for checking tags creation and deletion as per below comment.

pkg/resource/subnet/hooks.go
```diff
@@ -243,6 +250,131 @@ func toStrPtr(str string) *string {
...
+	if len(a.ko.Spec.Tags) != len(b.ko.Spec.Tags) {
+		delta.Add("Spec.Tags", a.ko.Spec.Tags, b.ko.Spec.Tags)
+	} else if len(a.ko.Spec.Tags) > 0 {
+		if !equalTags(a.ko.Spec.Tags, b.ko.Spec.Tags) {
+			delta.Add("Spec.Tags", a.ko.Spec.Tags, b.ko.Spec.Tags)
+		}
+	}
+}
+
+// equalTags function used to compare whether two tag arrays are equal or not
```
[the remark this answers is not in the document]
LikithaVemulapalli: Removed `equalTags` function, not required when using go templates.

brycahta (changes requested):
Can you regenerate without local changes in _code-generator_ to maintain `v0.19.3`? Otherwise, looks great!

templates/hooks/subnet/sdk_file_end.go.tpl
```diff
@@ -0,0 +1,32 @@
...
+
+func (rm *resourceManager) new{{ $subnetRefName }}(
+	    c svcapitypes.{{ $subnetRefName }},
+) *svcsdk.{{ $subnetRefName }} {
+	res := &svcsdk.{{ $subnetRefName }}{}
+{{ GoCodeSetSDKForStruct $CRD "" "res" $subnetRef "" "c" 1 }}
+	return res
+}
+{{- end }}
+{{- end }}
```
RedbackThomson, line 30: Ew. I hate that we have to do this. Not blaming you at all for this ugly mess, it totally makes sense. My s3 controller hacks unfortunately have set a precedent for relying too heavily on generated `sdk_file_end` hooks. @A-Hilaly I think we should expedite our generic tagging support to prevent ... this

A-Hilaly: +1 for what Nick said here. @RedbackThomson Indeed, i will open an issue to track and discuss how we can achieve the generic tagging support.

brycahta: Sounds good; lmk once you have the issue so we can gauge the effort. Might need to rely on copy/paste this template style for other resources if dev effort is more than a couple weeks

brycahta (approved):
/lgtm

PR #84, opened 2022-08-30
Refactor `VPCSpec.CIDRBlock` into `CIDRBlocks`

brycahta:
Issue #, if available: https://github.com/aws-controllers-k8s/community/issues/1198

Description of changes: VPCs can have multiple IPv4 CIDR blocks associated with it; therefore, refactor VPC's Spec `CIDRBlock` field into `CIDRBlocks` to support a list of CIDR blocks. The **first** entry in `VPCSpec.CIDRBlocks` will be used as the **primary** IPv4 CIDR block upon VPC creation. This is prerequisite work for `Update` logic and supporting multiple CIDR blocks will be enabled in a subsequent PR.


pkg/resource/vpc/hook.go
```diff
@@ -177,6 +177,16 @@ func (rm *resourceManager) customUpdate(
 	return &resource{ko}, nil
 }
 
+// applyPrimaryCIDRBlockInCreateRequest populates
+// CreateVpcInput.CidrBlock field with the FIRST
+// CIDR block defined in the resource's Spec
+func applyPrimaryCIDRBlockInCreateRequest(r *resource,
+	input *svcsdk.CreateVpcInput) {
+	if r.ko.Spec.CIDRBlocks != nil && len(r.ko.Spec.CIDRBlocks) > 0 {
```
A-Hilaly, line 185: nit: the nil check is not necessary - `len(r.ko.Spec.CIDRBlocks) > 0` should be enough.

config/crd/bases/ec2.services.k8s.aws_vpcs.yaml
```diff
@@ -92,8 +90,6 @@ spec:
                       type: string
                   type: object
                 type: array
-            required:
-            - cidrBlock
```
A-Hilaly, line 96: Idk if this will work, can we make `CIDRBlocks` required too?

A-Hilaly (approved):
👍 Left two comments that can be addressed in the following PR.
/lgtm

PR #85, opened 2022-08-31
Add `Update` functionality for `VPC.CIDRBlocks`

brycahta:
Issue #, if available: https://github.com/aws-controllers-k8s/community/issues/1198

Description of changes: Enables associating multiple IPv4 CIDR Blocks to a single VPC 

### DIFF
--- a/pkg/resource/vpc/hook.go
+++ b/pkg/resource/vpc/hook.go
@@ -168,11 +168,9 @@ func (rm *resourceManager) syncCIDRBlocks(
 	exit := rlog.Trace("rm.syncCIDRBlocks")
 	defer exit(err)
 
-	desRes := desired.ko.DeepCopy()
-	latestRes := latest.ko.DeepCopy()
-	desiredCIDRs := desRes.Spec.CIDRBlocks
-	latestCIDRs := latestRes.Spec.CIDRBlocks
-	latestCIDRStates := latestRes.Status.CIDRBlockAssociationSet
+	desiredCIDRs := desired.ko.Spec.CIDRBlocks
+	latestCIDRs := latest.ko.Spec.CIDRBlocks
+	latestCIDRStates := latest.ko.Status.CIDRBlockAssociationSet
 	toAddCIDRs, toDeleteCIDRs := computeStringPDifference(desiredCIDRs, latestCIDRs)
 
 	// extract associationID for the DisassociateVpcCidr request
--- a/pkg/resource/vpc/hook.go
+++ b/pkg/resource/vpc/hook.go
@@ -16,6 +16,7 @@ package vpc
 import (
 	"context"
 
+	svcapitypes "github.com/aws-controllers-k8s/ec2-controller/apis/v1alpha1"
 	ackcompare "github.com/aws-controllers-k8s/runtime/pkg/compare"
 	ackrtlog "github.com/aws-controllers-k8s/runtime/pkg/runtime/log"
 	svcsdk "github.com/aws/aws-sdk-go/service/ec2"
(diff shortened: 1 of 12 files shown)


pkg/resource/vpc/hook.go
```diff
@@ -128,6 +129,95 @@ func (rm *resourceManager) syncDNSHostnamesAttribute(
 	return nil
 }
 
+// computeStringPDifference uses the underlying string value
+// to discern which elements are in slice `a` that aren't in slice `b`
+// and vice-versa
+func computeStringPDifference(a, b []*string) (aNotB, bNotA []*string) {
```
brycahta, line 135: TODO: can make this generic and some type of common hook helper

pkg/resource/vpc/hook.go
```diff
@@ -128,6 +129,95 @@ func (rm *resourceManager) syncDNSHostnamesAttribute(
...
+	latestCIDRStates := latestRes.Status.CIDRBlockAssociationSet
+	toAddCIDRs, toDeleteCIDRs := computeStringPDifference(desiredCIDRs, latestCIDRs)
+
+	// extract associationID for the DisassociateVpcCidr request
+	for _, cidr := range toDeleteCIDRs {
+		input := &svcsdk.DisassociateVpcCidrBlockInput{}
+		for _, cidrAssociation := range latestCIDRStates {
+			if *cidr == *cidrAssociation.CIDRBlock {
+				input.AssociationId = cidrAssociation.AssociationID
+				_, err = rm.sdkapi.DisassociateVpcCidrBlockWithContext(ctx, input)
```
brycahta, line 184: Looking for input on how to test this (other than locally). I'm not able to remove a single entry from a list of entries via patch:

```
# context: current VPC has cidrBlocks X and Y
# this patch to remove Y does NOT work:
updates = {
            "spec": {"cidrBlocks": [X]}
        }

# this patch is used in other tests to delete a list completely, but cannot do that here because at least 1 cidr block is needed
updates = {
            "spec": {"cidrBlocks": []}
        }

```

pkg/resource/vpc/hook.go
A-Hilaly: Agreed. You can create a file under `pkg/util` and something like https://github.com/aws-controllers-k8s/rds-controller/blob/main/pkg/util/tags.go#L28. We should ideally move this kind of utility to github.com/aws-controllers-k8s/pkg

pkg/resource/vpc/hook.go
A-Hilaly: I run into something very similar previously: https://github.com/aws-controllers-k8s/lambda-controller/pull/39#discussion_r945228692
Maybe using `None` can help?

pkg/resource/vpc/hook.go
```diff
@@ -128,6 +129,95 @@ func (rm *resourceManager) syncDNSHostnamesAttribute(
...
+	latestRes := latest.ko.DeepCopy()
+	desiredCIDRs := desRes.Spec.CIDRBlocks
+	latestCIDRs := latestRes.Spec.CIDRBlocks
+	latestCIDRStates := latestRes.Status.CIDRBlockAssociationSet
+	toAddCIDRs, toDeleteCIDRs := computeStringPDifference(desiredCIDRs, latestCIDRs)
+
+	// extract associationID for the DisassociateVpcCidr request
+	for _, cidr := range toDeleteCIDRs {
+		input := &svcsdk.DisassociateVpcCidrBlockInput{}
+		for _, cidrAssociation := range latestCIDRStates {
```
A-Hilaly, line 181: Why not directly call `rm.sdkapi.DisassociateVpcCidrBlockWithContext(ctx, input)` for those, i believe the information returned by `computeStringPDifference` is correct.

pkg/resource/vpc/hook.go
```diff
@@ -128,6 +129,95 @@ func (rm *resourceManager) syncDNSHostnamesAttribute(
...
+	ctx context.Context,
+	desired *resource,
+	latest *resource,
+) (err error) {
+	rlog := ackrtlog.FromContext(ctx)
+	exit := rlog.Trace("rm.syncCIDRBlocks")
+	defer exit(err)
+
+	desRes := desired.ko.DeepCopy()
+	latestRes := latest.ko.DeepCopy()
```
A-Hilaly, line 172: nit: i think there no need to deep copying here, as there are no statements mutating `desired` and `latest` objects

pkg/resource/vpc/hook.go
```diff
@@ -128,6 +129,95 @@ func (rm *resourceManager) syncDNSHostnamesAttribute(
...
+			return err
+		}
+	}
+
+	return nil
+}
+
+// setSpecCIDRs sets Spec.CIDRBlocks using the CIDRs in
+// Status.CIDRBlockAssociationSet, which is set via sdkCreate/sdkFind
+func (rm *resourceManager) setSpecCIDRs(
```
A-Hilaly, line 210: +++

pkg/resource/vpc/hook.go
brycahta: By *for those* do you mean `toDeleteCIDRs` ?

The information returned by `computeStringPDifference` is correct, but only the _CIDR Blocks_. I need to extract the `AssociationID` from the currently existing CIDR Block associations because it's required for `DisassociateVpcCidrBlock` request

pkg/resource/vpc/hook.go
brycahta: Can I address in separate PR?

pkg/resource/vpc/hook.go
brycahta: I don't think I can use `None` because I need at least 1 CIDR Block and I'm trying to remove a single CIDR Block while the VPC is associated with multiple

pkg/resource/vpc/hook.go
A-Hilaly: Sure, feel free to add it when you think it's the right time.

pkg/resource/vpc/hook.go
A-Hilaly: Oh i get it now. Didn't know that the association ID is what's needed.

pkg/resource/vpc/hook.go
A-Hilaly: Just spoke with @RedbackThomson about this issue. For now we don't have a solution for it using our testing tools.

A-Hilaly (approved):

pkg/resource/vpc/hook.go
RedbackThomson: > I don't think I can use `None` because I need at least 1 CIDR Block and I'm trying to remove a single CIDR Block while the VPC is associated with multiple

`kubectl patch` does not support "updating" lists or dictionaries. The only way to achieve this is by describing the current list, modifying it within a python variable, then replacing the whole list. This should have the desired effect, though.

pkg/resource/vpc/hook.go
RedbackThomson: > Agreed. You can create a file under `pkg/util` and something like https://github.com/aws-controllers-k8s/rds-controller/blob/main/pkg/util/tags.go#L28. We should ideally move this kind of utility to github.com/aws-controllers-k8s/pkg

https://github.com/samber/lo 😄 Just need go1.18

brycahta: * Is there anything blocking us from upgrading to go1.18 today? Wouldn't be opposed to including it in a near-future release

pkg/resource/vpc/hook.go
brycahta: Thanks @RedbackThomson , let me give that a try!

PR #90, opened 2022-09-07
UpdateTags support for VPC

LikithaVemulapalli:
Description of changes:

- Implemented updateTags support for **VPC** resource using go templates and hook code.
- Added integration tests for updateTags.
- Removed existing `delta_pre_compare.go.tpl` file from subnet templates as we are using `sdk_file_end.go.tpl`.

### DIFF
--- a/pkg/resource/vpc/hook.go
+++ b/pkg/resource/vpc/hook.go
@@ -281,6 +281,121 @@ func applyPrimaryCIDRBlockInCreateRequest(r *resource,
 	}
 }
 
+// syncTags used to keep tags in sync using createTags and deleteTags functions
+func (rm *resourceManager) syncTags(
+	ctx context.Context,
+	desired *resource,
+	latest *resource,
+) (err error) {
+	rlog := ackrtlog.FromContext(ctx)
+	exit := rlog.Trace("rm.syncTags")
+	defer func(err error) {
+		exit(err)
+	}(err)
+
+	toAdd, toDelete := computeTagsDelta(
+		desired.ko.Spec.Tags, latest.ko.Spec.Tags,
+	)
+
+	if err = rm.deleteTags(ctx, latest, toDelete); err != nil {
+		return err
+	}
+
+	if err = rm.createTags(ctx, desired, toAdd); err != nil {
+		return err
+	}
+
+	return nil
+}
+
+// createTags function creates tags for VPC resource using CreateTags API calls
+func (rm *resourceManager) createTags(
+	ctx context.Context,
+	r *resource,
+	tags []*svcapitypes.Tag,
+) (err error) {
+	rlog := ackrtlog.FromContext(ctx)
+	exit := rlog.Trace("rm.createTags")
+	defer exit(err)
+
+	resourceId := []*string{r.ko.Status.VPCID}
+
+	for _, i := range tags {
+		toAdd := rm.newTag(*i)
+		req := &svcsdk.CreateTagsInput{
+			Resources: resourceId,
+			Tags:      []*svcsdk.Tag{toAdd},
+		}
+		_, err := rm.sdkapi.CreateTagsWithContext(ctx, req)
+		rm.metrics.RecordAPICall("CREATE", "CreateTags", err)
+		if err != nil {
+			return err
+		}
+	}
+	return err
+}
+
+// deleteTags function deletes tags from VPC resource using DeleteTags API calls
+func (rm *resourceManager) deleteTags(
+	ctx context.Context,
+	r *resource,
+	tags []*svcapitypes.Tag,
+) (err error) {
+	rlog := ackrtlog.FromContext(ctx)
+	exit := rlog.Trace("rm.deleteTags")
+	defer exit(err)
+
+	resourceId := []*string{r.ko.Status.VPCID}
+
+	for _, i := range tags {
+		toDelete := rm.newTag(*i)
+		req := &svcsdk.DeleteTagsInput{
+			Resources: resourceId,
+			Tags:      []*svcsdk.Tag{toDelete},
+		}
+		_, err = rm.sdkapi.DeleteTagsWithContext(ctx, req)
+		rm.metrics.RecordAPICall("DELETE", "DeleteTags", err)
+		if err != nil {
+			return err
+		}
+	}
+	return err
+}
+
+// computeTagsDelta returns tags to be added and removed from the resource
+func computeTagsDelta(
+	desired []*svcapitypes.Tag,
+	latest []*svcapitypes.Tag,
+) (toAdd []*svcapitypes.Tag, toDelete []*svcapitypes.Tag) {
+
+	desiredTags := map[string]string{}
+	for _, tag := range desired {
+		desiredTags[*tag.Key] = *tag.Value
+	}
+
+	latestTags := map[string]string{}
+	for _, tag := range latest {
+		latestTags[*tag.Key] = *tag.Value
+	}
+
+	for _, tag := range desired {
+		val, ok := latestTags[*tag.Key]
+		if !ok || val != *tag.Value {
+			toAdd = append(toAdd, tag)
+		}
+	}
+
+	for _, tag := range latest {
+		_, ok := desiredTags[*tag.Key]
+		if !ok {
+			toDelete = append(toDelete, tag)
+		}
+	}
+
+	return toAdd, toDelete
+
+}
+
 // updateTagSpecificationsInCreateRequest adds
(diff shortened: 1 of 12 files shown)


test/e2e/conftest.py
```diff
@@ -74,6 +74,10 @@ def simple_vpc(request):
             replacements["ENABLE_DNS_SUPPORT"] = data['enable_dns_support']
         if 'enable_dns_hostnames' in data:
             replacements["ENABLE_DNS_HOSTNAMES"] = data['enable_dns_hostnames']
+        if 'key' in data:
+            replacements["KEY"] = data['key']
+        if 'value' in data:
+            replacements["VALUE"] = data['value']
```
jaypipes, line 80: recommend naming these variables `TAG_KEY` and `TAG_VALUE` to save the reader some trouble in the future wondering what they refer to :)

templates/hooks/vpc/sdk_file_end.go.tpl
```diff
@@ -0,0 +1,32 @@
...
+{{/* Generate helper methods for VPC */}}
+{{- range $specFieldName, $specField := $CRD.Config.Resources.Vpc.Fields }}
+{{- if $specField.From }}
+{{- $operationName := $specField.From.Operation }}
+{{- $operation := (index $SDKAPI.API.Operations $operationName) -}}
+{{- range $vpcRefName, $vpcMemberRefs := $operation.InputRef.Shape.MemberRefs -}}
+{{- if eq $vpcRefName "Tags" }}
+{{- $vpcRef := $vpcMemberRefs.Shape.MemberRef }}
+{{- $vpcRefName = "Tag" }}
+func compare{{ $vpcRefName }}(
```
jaypipes, line 13: Where are any of these `compareXXX` functions being called?

LikithaVemulapalli: My bad! Initially I did implement using compareTags function, but later after suggestions in previous subnet PR, converted logic to use go templates, thought I would need `compareTag` at some point later and included it, cross-checked with local testing, no need for this `compareTag` function anymore, thank you for noticing it! Will remove this function for `subnet` resource as well!

test/e2e/conftest.py
```diff
@@ -74,6 +74,10 @@ def simple_vpc(request):
             replacements["ENABLE_DNS_SUPPORT"] = data['enable_dns_support']
         if 'enable_dns_hostnames' in data:
             replacements["ENABLE_DNS_HOSTNAMES"] = data['enable_dns_hostnames']
+        if 'key' in data:
```
brycahta, line 77: nit: could you update this to also be `tag_key` / `tag_value`?

brycahta (approved):
Thanks @LikithaVemulapalli ; looks good! Will let @jaypipes re-review before merging

A-Hilaly (approved):
:+1:

pkg/resource/vpc/hook.go
```diff
@@ -281,6 +287,121 @@ func applyPrimaryCIDRBlockInCreateRequest(r *resource,
...
+	r *resource,
+	tags []*svcapitypes.Tag,
+) (err error) {
+	rlog := ackrtlog.FromContext(ctx)
+	exit := rlog.Trace("rm.deleteTags")
+	defer exit(err)
+
+	resourceId := []*string{r.ko.Status.VPCID}
+
+	for _, i := range tags {
```
jaypipes, line 356: Is there a reason we need to make multiple calls to `DeleteTags` instead of just calling it once with the `DeleteTagsInput.Tags` field set to all the tags we want to delete?

pkg/resource/vpc/hook.go
```diff
@@ -281,6 +287,121 @@ func applyPrimaryCIDRBlockInCreateRequest(r *resource,
...
+	r *resource,
+	tags []*svcapitypes.Tag,
+) (err error) {
+	rlog := ackrtlog.FromContext(ctx)
+	exit := rlog.Trace("rm.createTags")
+	defer exit(err)
+
+	resourceId := []*string{r.ko.Status.VPCID}
+
+	for _, i := range tags {
```
jaypipes, line 329: Is there a reason we need to make multiple calls to `CreateTags` instead of just calling it once with the `CreateTagsInput.Tags` field set to all the tags we want to create?

pkg/resource/vpc/hook.go
LikithaVemulapalli: Initially I thought it would be clear to have separate functions for creation and deletion of tags. To align with the other svcapitypes to svcsdk tag conversions in EC2 controller, I need to use the go templates. The `newTag` function converts only single tag at a time (cross-checked with the target shape) , so thought about the iterations. But, this can be done in other way as well, like instead of making multiple calls to API's, we can add a function to convert to sdktags and call it in `DeleteTagsInput.Tags`, will change accordingly, if you feel this would be more ideal! Thank you!

pkg/resource/vpc/hook.go
LikithaVemulapalli: Same explanation as above! Will change it accordingly if needed!

jaypipes: Sorry @LikithaVemulapalli I'm not explaining myself correctly. I meant would it be possible to reduce the number of API calls made here (and below in the delete code paths) by doing this?

```go
func (rm *resourceManager) createTags(
	ctx context.Context,
	r *resource,
	tags []*svcapitypes.Tag,
) (err error) {
	rlog := ackrtlog.FromContext(ctx)
	exit := rlog.Trace("rm.createTags")
	defer exit(err)

	resourceId := []*string{r.ko.Status.VPCID}

	sdkTags := make([]*svcsdk.Tag, len(tags))
	for x, tag := range tags {
		toAdd := rm.newTag(*tag)
		sdkTags[x] = toAdd
	}
	req := &svcsdk.CreateTagsInput{
		Resources: resourceId,
		Tags: sdkTags,
	}
	_, err := rm.sdkapi.CreateTagsWithContext(ctx, req)
	rm.metrics.RecordAPICall("CREATE", "CreateTags", err)
	return err
}
```

jaypipes (changes requested):

pkg/resource/vpc/hook.go
LikithaVemulapalli: Yes, it's possible to reduce number of API calls by calling a separate function to append all the converted sdk tags and passing it as input! Will change accordingly! Thank you!

brycahta (approved):
/lgtm

appreciate the improvements 👍

pkg/resource/vpc/hook.go
```diff
@@ -299,73 +299,57 @@ func (rm *resourceManager) syncTags(
...
+
 	toAdd, toDelete := computeTagsDelta(
 		desired.ko.Spec.Tags, latest.ko.Spec.Tags,
 	)
 
-	if err = rm.deleteTags(ctx, latest, toDelete); err != nil {
-		return err
-	}
+	if len(toDelete) > 0 {
+		rlog.Debug("removing tags from parameter group", "tags", toDelete)
```
jaypipes, line 309: is this removing tags from parameter group? or VPC? :)

pkg/resource/vpc/hook.go
```diff
@@ -299,73 +299,57 @@ func (rm *resourceManager) syncTags(
...
-	for _, i := range tags {
-		toAdd := rm.newTag(*i)
-		req := &svcsdk.CreateTagsInput{
-			Resources: resourceId,
-			Tags:      []*svcsdk.Tag{toAdd},
-		}
-		_, err := rm.sdkapi.CreateTagsWithContext(ctx, req)
-		rm.metrics.RecordAPICall("CREATE", "CreateTags", err)
+	if len(toAdd) > 0 {
+		rlog.Debug("adding tags to parameter group", "tags", toAdd)
```
jaypipes, line 325: ditto :)

pkg/resource/vpc/hook.go
brycahta: fixed in follow-up PR: https://github.com/aws-controllers-k8s/ec2-controller/pull/92/files#diff-12b9f9c625e91af04b8e0bb655157fff51de25c8f72b891312a761b212d3e018R309

pkg/resource/vpc/hook.go
LikithaVemulapalli: Fixed in follow-up PR: https://github.com/aws-controllers-k8s/ec2-controller/pull/92/files#diff-12b9f9c625e91af04b8e0bb655157fff51de25c8f72b891312a761b212d3e018R309

PR #91, opened 2022-09-08
remove dev preview from olm file in preparation for GA

acornett21:
Issue #, if available:
N/A

Description of changes:
Removing dev preview to prepare for GA as well as adding in the missing kinds to the samples section, so the CSV is fully filled out.

### DIFF
--- a/olm/olmconfig.yaml
+++ b/olm/olmconfig.yaml
@@ -18,7 +18,7 @@ description: |-
 
 
   This controller is a component of the [AWS Controller for Kubernetes](https://github.com/aws/aws-controllers-k8s)
-  project. This project is currently in **developer preview**.
+  project.
 
 
   **Pre-Installation Steps**
@@ -39,6 +39,14 @@ samples:
 - kind: Subnet
   spec: '{}'
 - kind: TransitGateway
+  spec: '{}'
+- kind: Instance
+  spec: '{}'
+- kind: DHCPOptions
+  spec: '{}'
+- kind: NATGateway
+  spec: '{}'
+- kind: ElasticIPAddress
   spec: '{}'
 maintainers:
 - name: "ec2 maintainer team"


brycahta (changes requested):
Thanks for the PR @acornett21. Can we remove `Instance` from the list? We are going GA with the network-related resources first: https://github.com/aws-controllers-k8s/community/issues/489  Everything else looks good

acornett21:
> Can we remove `Instance` from the list? 
@brycahta How are you picking and choosing which resources? `Instance` is part of the codebase, so when everything is generated this CRD will be a resource that is packaged, and will show as an option for users to create a CR from.

brycahta:
> > Can we remove `Instance` from the list?
> 
> @brycahta How are you picking and choosing which resources? `Instance` is part of the codebase, so when everything is generated this CRD will be a resource that is packaged, and will show as an option for users to create a CR from.

The linked issue lists the initial resources for ec2-controller:

1) [VPC](https://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_CreateVpc.html)
1) [Subnet](https://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_CreateSubnet.html)
1) [RouteTable](https://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_CreateRouteTable.html)
1) [Route](https://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_CreateRoute.html)
1) [InternetGateway](https://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_CreateInternetGateway.html)
1) [TransitGateway](https://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_CreateTransitGateway.html)
1) [VPC Endpoint](https://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_CreateVpcEndpoint.html)
1) [DHCP Options](https://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_CreateDhcpOptions.html)
1) [SecurityGroup](https://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_CreateSecurityGroup.html)


I see your point about `Instance` being in the codebase already; I had added that as a "Preview" while blocked on other work. Could you generate this file and remove `Instance` locally?

acornett21:
@brycahta I can exclude `Instance` from this file, but that doesn't solve the real issue. The real issue is that `Instance` is the codebase, and when the release process runs, this resource will be part of that process and `Instance` will be part of both the `helm` install (I don't know the right word for this), and it will be part of the Operator Bundle targeting OLM. So the intent might be to exclude this resource from GA, but it will be shipped to customers anyways. 

@jaypipes Any thoughts on how we want to handle this resources that are not meant for GA but are already part of the codebase?

acornett21:
I'll also note that this has already been released with `Instance` in the codebase so instance already exists in clusters, so removing it now would be problematic. The removal of this resource would require manual cluster admin intervention since OLM does not remove CRD's that have already been installed. See [here](https://github.com/k8s-operatorhub/community-operators/blob/main/operators/ack-ec2-controller/0.0.18/manifests/ec2.services.k8s.aws_instances.yaml)

brycahta:
> I'll also note that this has already been released with Instance in the codebase so instance already exists in clusters, so 
  removing it now would be problematic. The removal of this resource would require manual cluster admin intervention since 
  OLM does not remove CRD's that have already been installed. See [here](https://github.com/k8s-operatorhub/community- 
 operators/blob/main/operators/ack-ec2-controller/0.0.18/manifests/ec2.services.k8s.aws_instances.yaml)

@acornett21 appreciate the context I think I'm getting confused now. _ec2-controller_ is currently in [Preview](https://aws-controllers-k8s.github.io/community/docs/community/releases/#preview).[Released](https://aws-controllers-k8s.github.io/community/docs/community/releases/#released) stage-- doesn't this mean that we can freely change _ec2-controller_ including removing a CRD? From the [ACK Release documentation](https://aws-controllers-k8s.github.io/community/docs/community/releases/#released):

> Once a service controller reaches the RELEASED project stage, that does not mean that there can never be any changes 
  or additions to the Custom Resource Definitions (CRDs) or public interfaces exposed by that service controller. The 
  RELEASED project stage is simply an indication that there is at least one SemVer-tagged binary release of the controller.

I'll look into the effort involved to include `Instance` in GA, but also interested to hear @jaypipes thoughts.

jaypipes:
> > Once a service controller reaches the RELEASED project stage, that does not mean that there can never be any changes
> > or additions to the Custom Resource Definitions (CRDs) or public interfaces exposed by that service controller. The
> > RELEASED project stage is simply an indication that there is at least one SemVer-tagged binary release of the controller.
> 
> I'll look into the effort involved to include `Instance` in GA, but also interested to hear @jaypipes thoughts.

I would certainly prefer *not* to remove `Instance` from the list of supported resources, however I think adding some release notes stating that the `Instance` resource is experimental is fine.

We removed the `DBSecurityGroup` resource in the RDS controller but only because the `DBSecurityGroup` resource referred to now-deprecated Classic EC2 environments... so this is a different case.

brycahta (commented):
/ok-to-test

brycahta (approved):
/lgtm

acornett21:
@brycahta @jaypipes Thanks for the discussion and getting this merged!

PR #101, opened 2022-09-16
`Update` functionality for `Subnet` settings

brycahta:
Issue #, if available: https://github.com/aws-controllers-k8s/community/issues/1203

Description of changes: Updates `Subnet.Spec` fields, hooks, templates, and tests to align with `Subnet > Edit subnet settings` console functionality


brycahta:
> @brycahta: The following test **failed**, say `/retest` to rerun all failed tests:
> 
> Test name	Commit	Details	Rerun command
> ec2-kind-e2e	[e57affe](https://github.com/aws-controllers-k8s/ec2-controller/commit/e57affef719a893f7ccd5d2cc97068f6a77bfc2d)	[link](https://prow.ack.aws.dev/view/s3/ack-prow-logs/pr-logs/pull/aws-controllers-k8s_ec2-controller/101/ec2-kind-e2e/1570807709014429696)	`/test ec2-kind-e2e`
> [Full PR test history](https://prow.ack.aws.dev/pr-history?org=aws-controllers-k8s&repo=ec2-controller&pr=101). [Your PR dashboard](https://prow.ack.aws.dev/pr?query=is:pr+state:open+author:brycahta).

**root cause** is `Subnet::test_crud_tags` test failing most likely due to this [change](https://github.com/aws-controllers-k8s/ec2-controller/pull/101/files#diff-30a25a7dce40666fdab92a1b991943d7b74d0e47d4bcb4a4e1ebe74cdb93e00cL53) which no longer returns the `desired` resource, but the `latest` resource after a call to `sdkFind`. 

The call to `sdkFind` is necessary to update `Subnet.Status` fields following the `updateSubnetAttribute` call(s). I think this is the correct logic, so I'll try to update the test using a python variable to define empty tags list.

brycahta:
`RouteTable::test_crud_route` and `Instance:: test_crud_tags ` are failing now despite their code being untouched. Looks related to patching empty list again, but not sure why it's failing now? My tests are passing locally so not sure if it's a timing issue or possibly transient. Continuing to deep dive

generator.yaml
```diff
@@ -402,6 +402,20 @@ resources:
...
+      EnableDns64:
+        type: bool
+      EnableResourceNameDnsARecord:
+        type: bool
+      EnableResourceNameDnsAAAARecord:
+        type: bool
+      HostnameType:
+        type: string
+      MapPublicIpOnLaunch:
+        type: bool
```
jaypipes, line 418: @brycahta so did you have to put these fields into the `generator.yaml` file for a specific reason? Were they not being picked up automatically by the code generator during API inference stage?

templates/hooks/subnet/sdk_create_post_set_output.go.tpl
```diff
@@ -1,3 +1,19 @@
...
+	}
+	if desired.ko.Spec.CustomerOwnedIPv4Pool != nil {
+		ko.Spec.CustomerOwnedIPv4Pool = desired.ko.Spec.CustomerOwnedIPv4Pool
+	}
+	if desired.ko.Spec.EnableDNS64 != nil {
+		ko.Spec.EnableDNS64 = desired.ko.Spec.EnableDNS64
+	}
+	if desired.ko.Spec.MapPublicIPOnLaunch != nil {
+		ko.Spec.MapPublicIPOnLaunch = desired.ko.Spec.MapPublicIPOnLaunch
+	}
```
jaypipes, line 15: Any reason not to include `EnableResourceNameDNSAAAARecord`, `EnableResourceNameDNSARecord` or `HostnameType` in this hook?

pkg/resource/subnet/sdk.go
```diff
@@ -463,6 +475,22 @@ func (rm *resourceManager) sdkCreate(
...
+	if desired.ko.Spec.CustomerOwnedIPv4Pool != nil {
+		ko.Spec.CustomerOwnedIPv4Pool = desired.ko.Spec.CustomerOwnedIPv4Pool
+	}
+	if desired.ko.Spec.EnableDNS64 != nil {
+		ko.Spec.EnableDNS64 = desired.ko.Spec.EnableDNS64
+	}
+	if desired.ko.Spec.MapPublicIPOnLaunch != nil {
+		ko.Spec.MapPublicIPOnLaunch = desired.ko.Spec.MapPublicIPOnLaunch
+	}
+
```
jaypipes, line 493: Are you saying that all of these fields have incorrect values returned by the Create Output shape?

generator.yaml
brycahta: The fields were getting picked up by code-generator; however, because they were in the _Describe_ output shape, they were inferred as `Status` fields. I needed this config to move these fields to `Spec` because they can be edited by the user. I will add a comment in the next revision to make this clearer.

Note, I did try this config because I think it conveys the intent clearer, but it did not work:

```
MapPublicIpOnLaunch:
  is_read_only: false
```

Let me know if you think it's worth opening an issue

templates/hooks/subnet/sdk_create_post_set_output.go.tpl
brycahta: Yeah, these fields start off with `desired` values and the values will never be overridden/assigned by the response because the fields aren't present in the output shape (at the same level), so the resource data should remain accurate.

generator.yaml
jaypipes: @brycahta sorry for delay in response. What you could do here instead is do this:

```yaml
Subnet:
  fields:
    AssignIPV6AddressOnCreation:
      from:
        operation: DescribeSubnets
        path: AssignIpv6AddressOnCreation
```

The benefit of that is that the field's API documentation string will remain with the CRD's Field.

brycahta: I tried this config, but the fields will NOT be relocated to `Spec` but remain in `Status`.

jaypipes (approved):
Rock on, thanks @brycahta!

PR #102, opened 2022-09-16
Update tags support for NAT gateway

LikithaVemulapalli:
Description of changes:

- Included `update_operation` in `generator.yaml` file.
- Implemented update tags support for **NATGateway** resource.
- Added e2e tests for update tags.

### DIFF
--- a/pkg/resource/nat_gateway/hook.go
+++ b/pkg/resource/nat_gateway/hook.go
@@ -14,8 +14,136 @@
 package nat_gateway
 
 import (
+	"context"
+
+	svcapitypes "github.com/aws-controllers-k8s/ec2-controller/apis/v1alpha1"
+	ackcompare "github.com/aws-controllers-k8s/runtime/pkg/compare"
+	ackrtlog "github.com/aws-controllers-k8s/runtime/pkg/runtime/log"
 	svcsdk "github.com/aws/aws-sdk-go/service/ec2"
 )
+
+func (rm *resourceManager) customUpdateNATGateway(
+	ctx context.Context,
+	desired *resource,
+	latest *resource,
+	delta *ackcompare.Delta,
+) (updated *resource, err error) {
+	rlog := ackrtlog.FromContext(ctx)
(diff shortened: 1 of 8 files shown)


pkg/resource/nat_gateway/hook.go
```diff
@@ -14,9 +14,137 @@
...
+	ko := desired.ko.DeepCopy()
+
+	if delta.DifferentAt("Spec.Tags") {
+		if err := rm.syncTags(ctx, desired, latest); err != nil {
+			return nil, err
+		}
+	}
+
+	rm.setStatusDefaults(ko)
+	return &resource{ko}, nil
```
brycahta, line 46: @A-Hilaly @RedbackThomson 

Can you help me understand the expected behavior here? The current code will return the `desired` resource and therefore `desired.Spec.Tags` **without** checking server-side values. As long as there's no error from `syncTags`, then controller will see the update as successful because there will be no difference between the _desired_ and _returned_ resource tags. I'm not sure if this is a safe assumption given some tags can be added to the resource (i.e. ACK tags via `ensureTags` or tags via Console).  _Note, if server-side tag values exist that are not present in Spec (i.e. ACK tags) then they will NOT be present in `Spec` until next `sdkFind`_

For additional context, in other resources such as `SecurityGroup` , `RouteTable`, `Subnet` there's a call to `sdkFind` (within `sdkUpdate`) to "refresh" the resource's Spec/Status fields with the latest data prior to returning resource. The logic being `sdkUpdate` returns the resource's state after it has been updated-- is this correct?

A-Hilaly: Your assumption is correct @brycahta, since you are only updating tags and tags update happens instantly.

If we assume that fields are instantly updated in the backend API, we can return the desired object with no issues. In case where things can happen asynchronously, we can execute a ReadOne call and return the result instead: [example in lambda-controller](https://github.com/aws-controllers-k8s/lambda-controller/blob/main/pkg/resource/function/hooks.go#L115-L119)

brycahta (approved):
/lgtm

PR #103, opened 2022-09-19
Update tags support for RouteTable resource

LikithaVemulapalli:
Description of changes:

- Implemented `updateTags` support for **RouteTable** resource.
- Added e2e tests for update tags.


LikithaVemulapalli:
Modified `RouteTable` and `Instance` tags support tests as the tests are failing on checking whether tags are deleted or not. When the resource is created, along with the tags that we specify, two service tags are also getting added. When we delete the tags from resource, only the specified tags are removed but not the service tags, modified the test to check if the specified tags are removed or not from the tag list.

brycahta (approved):
/lgtm

Thanks for tag support @LikithaVemulapalli

PR #105, opened 2022-09-26
Refactor `customUpdate` hooks

brycahta:
Issue #, if available: N/A

Description of changes: The goal of this refactor is to align `customUpdate` hook code across resources so the codebase is easier to read and maintain. Changes to make the code more consistent include:
* rename `hook.go` to `hooks.go`
* return *updated* resource in `customUpdate` with comments
* elevating API calls to `customUpdate` (i.e. moving `addRulesToStatus` out of `sync`)
* rename methods to be more explicit about underlying API calls (i.e. `addRulesToStatus` --> `sdkFindRules`)


pkg/resource/dhcp_options/hooks.go
```diff
@@ -32,18 +32,20 @@ func (rm *resourceManager) customUpdateDHCPOptions(
...
+	// (now updated.Spec) reflects the latest resource state.
+	updated = desired
 
 	if delta.DifferentAt("Spec.Tags") {
 		if err := rm.syncTags(ctx, desired, latest); err != nil {
 			return nil, err
 		}
 	}
 
-	rm.setStatusDefaults(ko)
```
jaypipes, line 45: Any reason you're removing this?

If you simply removed the names of the return variables, and didn't change anything else about this method, I believe you would be OK...

pkg/resource/security_group/hooks.go
```diff
@@ -88,11 +90,11 @@ func (rm *resourceManager) addRulesToStatus(
...
 	}
 	if err != nil {
-		return err
+		return nil, err
 	}
 
-	ko.Status.Rules = rulesForResource
-	return nil
+	latest.ko.Status.Rules = rulesForResource
+	return latest, nil
```
jaypipes, line 97: Generally I prefer *not* to set CR Spec or Status values from within an internal hook function/method.

Instead, I prefer to isolate the "getter" method to just return the value that will be set in the Spec or Status and then have `sdk_read_one_post_set_output` hook do the setting of Spec or Status to the returned value.

For instance, here is where the IAM controller's Role resource hooks implements the "getter" for a role's Policiies:

https://github.com/aws-controllers-k8s/iam-controller/blob/687eb2d43958f923a3691773bae262f38e1b08cc/pkg/resource/role/hooks.go#L74-L101

And here is where the hook code sets the Role's `Spec.Policies` value:

https://github.com/aws-controllers-k8s/iam-controller/blob/687eb2d43958f923a3691773bae262f38e1b08cc/templates/hooks/role/sdk_read_one_post_set_output.go.tpl#L8-L12

I find this pattern cleaner... make the method do one thing and do it well. In this case, the method should just return the rules for the security group. Rely on another method (in this case, the `sdkFind` method that has an `sdk_read_one_post_set_output` hook point for precisely this type of action.

pkg/resource/dhcp_options/hooks.go
brycahta: Yeah, I removed because I think the code is clearer when using the return variable, `updated`. 

The `ko` initialization and `rm.setStatusDefaults(ko)` make sense with `Create` and `ReadOne` because there's data in both responses. However, since EC2 `modify` APIs lack data (up to this point), I think reusing this same pattern can only introduce confusion. For example, initializing `ko := desired.ko.DeepCopy()` makes sense, but I don't think `rm.setStatusDefaults(ko)` will ever do anything because by the time we enter `customUpdate` the `desired` resource should be valid with its Status fields initialized already.

pkg/resource/security_group/hooks.go
brycahta: > I prefer to isolate the "getter" method to just return the value that will be set in the Spec or Status

I can refactor to align with this.

> then have `sdk_read_one_post_set_output` hook do the setting of Spec or Status to the returned value.

Hmm I don't understand why we'd always couple with the `sdkFind` hook instead of invoking the method when it's needed (i.e. in `customUpdate`). For `SecurityGroup` specifically, there would be 2 concerns:
1. `sdkFind` will always call `DescribeSecurityGroups` and `DescribeSecurityGroupRules` if invoked by `sdkUpdate`; however, only the latter call is needed
2. multiple `Spec` fields can be assigned in [`sdkFind` ](https://github.com/aws-controllers-k8s/ec2-controller/blob/main/pkg/resource/security_group/sdk.go#L55) such as `Tags`, `IngressRules`, and `EgressRules`. If we always call `sdkFind` with `sdkUpdate`, then it is _possible_ the `Spec` of the returned updated resource gets changed which wouldn't be expected after calling `sdkUpdate`

jaypipes: > > I prefer to isolate the "getter" method to just return the value that will be set in the Spec or Status
> 
> I can refactor to align with this.
> 
> > then have `sdk_read_one_post_set_output` hook do the setting of Spec or Status to the returned value.
> 
> Hmm I don't understand why we'd always couple with the `sdkFind` hook instead of invoking the method when it's needed (i.e. in `customUpdate`). For `SecurityGroup` specifically, there would be 2 concerns:
> 
> 1. `sdkFind` will always call `DescribeSecurityGroups` and `DescribeSecurityGroupRules` if invoked by `sdkUpdate`; however, only the latter call is needed
> 2. multiple `Spec` fields can be assigned in [`sdkFind` ](https://github.com/aws-controllers-k8s/ec2-controller/blob/main/pkg/resource/security_group/sdk.go#L55) such as `Tags`, `IngressRules`, and `EgressRules`. If we always call `sdkFind` with `sdkUpdate`, then it is _possible_ the `Spec` of the returned updated resource gets changed which wouldn't be expected after calling `sdkUpdate`

Unless the API call to create or update a resource is expected to change the value of a Spec field, there's no reason to set the value of a Spec field on the `latest` variable at all, though. Because, by default, `ko` (which ends up being returned in default sdkUpdate and sdkCreate calls is always set to a copy of `desired`, which already has those fields properly set...

brycahta: Now I think I'm confused. I agree with what you're saying which is why I set `updated = desired` in `customUpdate` with a comment explaining the same. Update is never expected to change the value of a Spec field; however, if a call to `sdkFind` is within Update, then it becomes possible that the `Spec` does get set before returning from Update. Because of this, I don't think we should always couple "getter" hook code with `sdkFind`

jaypipes (approved):
:+1:

brycahta:
/retest

needed to cleanup account

PR #106, opened 2022-09-27
Refactor e2e tests to use `acktest.tags` helper

brycahta:
Issue #, if available: https://github.com/aws-controllers-k8s/community/issues/1235

Description of changes: 
* Use `acktest.tags` assertions in e2e tests to accurately test for ACK tags and user-defined tags
* Enhance existing tag tests by checking both server-side values as well as `Spec.Tags` values
* Add hook code for `Instance` and `RouteTable` so ACK tags are not populated in `Spec`

### DIFF
--- a/templates/hooks/instance/sdk_create_post_set_output.go.tpl
+++ b/templates/hooks/instance/sdk_create_post_set_output.go.tpl
@@ -1,0 +1,6 @@
+	toAdd, toDelete := computeTagsDelta(desired.ko.Spec.Tags, ko.Spec.Tags)
+	if len(toAdd) == 0 && len(toDelete) == 0 {
+		// if desired tags and response tags are equal,
+		// then assign desired tags to maintain tag order
+		ko.Spec.Tags = desired.ko.Spec.Tags
+	}
--- a/test/e2e/requirements.txt
+++ b/test/e2e/requirements.txt
@@ -1,1 +1,1 @@
-acktest @ git+https://github.com/aws-controllers-k8s/test-infra.git@5220331d003e3a23de4470e68d02c99b16c81989
+acktest @ git+https://github.com/aws-controllers-k8s/test-infra.git@73eb0d5af8ac19d17fad8787d6462d56f4e3e37d
--- a/apis/v1alpha1/ack-generate-metadata.yaml
+++ b/apis/v1alpha1/ack-generate-metadata.yaml
@@ -1,5 +1,5 @@
 ack_generate_info:
-  build_date: "2022-09-21T13:36:34Z"
+  build_date: "2022-09-27T22:29:46Z"
   build_hash: 6e2ffbc3b16a30ac59be6719918c601c2c864064
   go_version: go1.18.1
   version: v0.20.1-3-g6e2ffbc
@@ -7,7 +7,7 @@ api_directory_checksum: 127aa0f51fbcdde596b8f42f93bcdf3b6dee8488
 api_version: v1alpha1
 aws_sdk_go_version: v1.44.93
 generator_config_info:
-  file_checksum: 8573a89b8220023263610579bfb84b762d99f5fa
+  file_checksum: a31177552ce6ea5ce3b624520711382f113cf05b
   original_file_name: generator.yaml
 last_modification:
   reason: API generation
--- a/apis/v1alpha1/generator.yaml
+++ b/apis/v1alpha1/generator.yaml
@@ -205,6 +205,10 @@ resources:
     hooks:
       sdk_create_post_build_request:
         template_path: hooks/instance/sdk_create_post_build_request.go.tpl
+      sdk_create_post_set_output:
+        template_path: hooks/instance/sdk_create_post_set_output.go.tpl
+      sdk_read_many_post_set_output:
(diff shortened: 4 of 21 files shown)


test/e2e/requirements.txt
```diff
@@ -1 +1 @@
-acktest @ git+https://github.com/aws-controllers-k8s/test-infra.git@5220331d003e3a23de4470e68d02c99b16c81989
\ No newline at end of file
+acktest @ git+https://github.com/aws-controllers-k8s/test-infra.git@73eb0d5af8ac19d17fad8787d6462d56f4e3e37d
```
brycahta, line 1: Needed to bring in an up-to-date `test-infra` so that `acktest.tags` would be available

templates/hooks/instance/sdk_create_post_set_output.go.tpl
```diff
@@ -0,0 +1,6 @@
+	toAdd, toDelete := computeTagsDelta(desired.ko.Spec.Tags, ko.Spec.Tags)
```
brycahta, line 1: Not proud of this hack, but wanted to get something out here working so we can discuss in detail. Without this hook code for `Instance` and `RouteTable`, their respective tag e2e test will fail because the `Spec` will contain a total of 3 tags (user-defined + ACK) instead of just 1 (user-defined).

After some deep dive, I believe the issue is the tags from an sdk response may be in a different order than the tags in the desired resource. For example, within `runtime::createResource` the call to create is made:
 [`latest, err = rm.Create(ctx, desired)`](https://github.com/aws-controllers-k8s/runtime/blob/ae946aa1e0dd309db0064c2c347b724074725c60/pkg/runtime/reconciler.go#L414)
 
When deep diving `Instance` I found that the `latest.ko.Spec.Tags` set after `rm.Create` will always be in a different order than `desired.ko.Spec.Tags` (this is NOT the case for other resources-- the order usually matches). Because these do not match, the subsequent call to `patchResourceMetadataAndSpec` will detect this difference and patch the `latest` resource with the _entire_ list of tags returned from `rm.Create`. I _think_ this is because of the [`patch := client.MergeFrom(dobj)`](https://github.com/aws-controllers-k8s/runtime/blob/ae946aa1e0dd309db0064c2c347b724074725c60/pkg/runtime/reconciler.go#L652) call within `patchResourceMetadataAndSpec`  which completely replaces new lists. From the [docs](https://kubernetes.io/docs/tasks/manage-kubernetes-objects/update-api-object-kubectl-patch/):

> When using MergeFrom, existing lists will be completely replaced by new lists.

vijtrip2: Hey Bryan, thanks for the deep dive on this.

I believe we can solve this problem at the root level. Can we look into having the tags in sorted order inside desired and latest resource from ACK runtime and that way individual teams won't have to add hooks like you did in this PR.

If you agree with my proposal, please file an issue in the community repo for it. And it is not blocking for this PR.

brycahta: https://github.com/aws-controllers-k8s/community/issues/1492

A-Hilaly: Just chatted with @brycahta and looks like we have a solution that will work without making changes to runtime. 
We have seen similar issues with others resources in different controllers when it comes to array type **fields** (example `[]*struct{Key, Value string}`). The problem here is more related to the generated code in `delta.go`/`rm.compareResources` -  the generated code uses `DeepEqual` from the  `reflect` library which takes into consideration the order of the elements, quoting:

```go
// Slice values are deeply equal when all of the following are true:
// they are both nil or both non-nil, they have the same length,
// and either they point to the same initial entry of the same underlying array
// (that is, &x[0] == &y[0]) or their corresponding elements (up to length) are deeply equal.
// Note that a non-nil empty slice and a nil slice (for example, []byte{} and []byte(nil))
// are not deeply equal.
```

On of the workarounds we used in lambda/sfn/AMP controllers was to inject a `preCompare` hook to properly compare arrays regardless of their order + instructing the code-generator to not generated comparison code for array type fields. Using the previous will not trigger the update calls for resources and will also prevent the reconcile function from overriding new tags to the Spec.

Example for tags in sfn: https://github.com/aws-controllers-k8s/sfn-controller/blob/main/pkg/resource/activity/hooks.go#L74
Another example in lambda for an array type field: https://github.com/aws-controllers-k8s/lambda-controller/pull/39/files

brycahta: I think this is a clearer workaround than what I've done. I can implement for all resources unless there's another a suggestion. 

Long term, I'm not sure if runtime is the place to implement a fix. Since this is needing to be done for multiple service repos and resources contained within, I think code-generator should be addressing it.

vijtrip2: Discussed with Amine offline. Amine solution works but i had some concerns for TagOnCreate with lateInitialize method. He will test with lambda controller locally and handle it while working on issue #1492.

This PR can be merged to unblock EC2 controller.

vijtrip2 (commented):
/lgtm

PR #107, opened 2022-10-03
Set `Instance` to synced iff it's in a permanent state

brycahta:
**Issue #, if available:** N/A. `Instance` would be set to `Synced` immediately after creation while it is still in 'pending' state. Because it's `Synced`, reconciliation ceases and the `Instance` state is never updated to 'running' (or any other permanent/terminal state).

**Description of changes:** Re-queue `Instance` if it is in a temporary state (pending, shutting-down, stopping) until a permanent state is achieved; update tests.

### DIFF
--- a/apis/v1alpha1/ack-generate-metadata.yaml
+++ b/apis/v1alpha1/ack-generate-metadata.yaml
@@ -1,5 +1,5 @@
 ack_generate_info:
-  build_date: "2022-09-27T22:29:46Z"
+  build_date: "2022-10-03T20:46:38Z"
   build_hash: 6e2ffbc3b16a30ac59be6719918c601c2c864064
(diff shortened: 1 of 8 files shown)


jaypipes (approved):

brycahta:
Expecting **e2e test failure**
* logs:
```
        # Ensure server-side state is propogated to the resource
        instance_cr = k8s.get_resource(ref)
>       assert instance_cr["status"]["state"]["name"] == "running"
[1m[31mE       AssertionError: assert 'pending' == 'running
```
* the resource is getting to synced state, but its `Status` is not being updated properly

From local testing, I think the **root cause** may be with **late initialization** code. Discussing with @vijtrip2 offline to confirm

PR #109, opened 2022-10-11
Fix hook code for `Tags`

brycahta:
Issue #, if available: https://github.com/aws-controllers-k8s/community/issues/1493

Description of changes:
* set `input.TagSpecifications` to non-nil in `updateTagSpecificationsInCreateRequest` iff there are desired tags
* apply workaround to all resources to maintain `Spec.Tags` order; otherwise, unnecessary `Update` calls may be triggered

### DIFF
--- a/apis/v1alpha1/ack-generate-metadata.yaml
+++ b/apis/v1alpha1/ack-generate-metadata.yaml
@@ -1,13 +1,13 @@
 ack_generate_info:
-  build_date: "2022-10-12T21:27:52Z"
+  build_date: "2022-10-13T18:47:02Z"
   build_hash: 5ee0ac052c54f008dff50f6f5ebb73f2cf3a0bd7
-  go_version: go1.18.2
+  go_version: go1.18.1
   version: v0.20.1-4-g5ee0ac0
-api_directory_checksum: 3638dfc2a8e94a1ea91486172874d0b0b4e039f7
+api_directory_checksum: b3a2878ca8a156389214b900257c4d572ad4e3a5
 api_version: v1alpha1
 aws_sdk_go_version: v1.44.93
 generator_config_info:
-  file_checksum: b50a20e57aaf81d2a636ebf9a74d9693a82eeeb7
+  file_checksum: 0260a4e58e3e1667fc456baff448cbabc2da4bf6
   original_file_name: generator.yaml
 last_modification:
   reason: API generation
--- a/apis/v1alpha1/generator.yaml
+++ b/apis/v1alpha1/generator.yaml
@@ -185,6 +185,8 @@ resources:
           operation: CreateTags
           path: Tags
     hooks:
+      delta_pre_compare:
+        code: compareTags(delta, a, b)
       sdk_create_post_build_request:
         template_path: hooks/dhcp_options/sdk_create_post_build_request.go.tpl
       sdk_file_end:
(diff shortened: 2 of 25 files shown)


brycahta:
Account is clean; failure seems transient:

```
    @pytest.fixture
    def simple_nat_gateway(standard_elastic_address, request):
        test_resource_values = REPLACEMENT_VALUES.copy()
        resource_name = random_suffix_name("nat-gateway-test", 24)
        test_vpc = get_bootstrap_resources().SharedTestVPC
        subnet_id = test_vpc.public_subnets.subnet_ids[0]
    
        (_, eip) = standard_elastic_address
    
        test_resource_values["NAT_GATEWAY_NAME"] = resource_name
        test_resource_values["SUBNET_ID"] = subnet_id
>       test_resource_values["ALLOCATION_ID"] = eip["status"]["allocationID"]
KeyError: 'allocationID'
```

Re-testing

jaypipes (commented):
@brycahta it seems to me that we need to fix the `Delta` comparison code for Tags in EC2 controller. That is the root cause of these issues I believe, and if we fix that, you won't need to do things like:

```go
	toAdd, toDelete := computeTagsDelta(r.ko.Spec.Tags, ko.Spec.Tags)
	if len(toAdd) == 0 && len(toDelete) == 0 {
		// if resource's initial tags and response tags are equal,
		// then assign resource's tags to maintain tag order
		ko.Spec.Tags = r.ko.Spec.Tags
	}
```

at all in the `sdkFind` method. Instead, you will simply be able to do this in your `sdkUpdate` method:

```go
if delta.DifferentAt("Spec.Tags") {
    if err := rm.SyncTags(...); err != nil {
        return nil, err
    }
}
```

brycahta:
@jaypipes I don't disagree with fixing the issue in `Delta` code. For context, I went with this workaround because it would solve the comparison issue AND return the resource with the Tags listed in the same order as they were defined. Do we need to persist the order of `Tags` (or any list field) or is that not important?

**Detailed Details**
If we implement the fix in `Delta` code only, then it's possible we return the resource with Tags in a different order than what was specified during `runtime::createResource`:

[code](https://github.com/aws-controllers-k8s/runtime/blob/13496a878abc8dd9f784d84c6452638a0d736df3/pkg/runtime/reconciler.go#L447)
```
      // desired tags: [A, B, C]  --> entered by User
      
       latest, err = rm.Create(ctx, desired)
      // latest tags: [B, A, C] --> eventually returned to User
       ...
	err = r.patchResourceMetadataAndSpec(ctx, desired, latest)
	if err != nil {
		return latest, err
	}
	rlog.Info("created new resource")
	return latest, nil
```



 If it's not important to persist the order as it was defined in the `Spec`, then I can move all of this into `Delta` hooks.

jaypipes:
> If it's not important to persist the order as it was defined in the `Spec`, then I can move all of this into `Delta` hooks.

Yeah, I don't think it's important to persist the order as it was defined in the `Spec`, since there's nothing significant about that order. Fixing it in the Delta code has the additional benefit that the logs for the controller won't fill up with needless `differences detected in Spec` ... messages when the only difference is the order of the tags.

brycahta:
@jaypipes , got it I can move to `Delta`.

> Fixing it in the Delta code has the additional benefit that the logs for the controller won't fill up with needless differences detected in Spec ... messages when the only difference is the order of the tags.

Just a note, the current implementation also does not fill up the logs because differences are never detected since order is enforced.

jaypipes:
> @jaypipes , got it I can move to `Delta`.
> 
> > Fixing it in the Delta code has the additional benefit that the logs for the controller won't fill up with needless differences detected in Spec ... messages when the only difference is the order of the tags.
> 
> Just a note, the current implementation also does not fill up the logs because differences are never detected since order is enforced.

Ah, yes, good point.

jaypipes (approved):
perfect. thanks @brycahta!

PR #110, opened 2022-10-12
enhance output with additionalPrinterColumns

smcavallo:
Signed-off-by: smcavallo <smcavallo@hotmail.com>

Issue #, if available:

Description of changes:
Adds ID and STATE output to kubectl for several resource types.
These are useful in general, and especially useful when creating a new resource to identify something which can't be successfully created or needs to be investigated.

This change should purely be kubegen/output only.

```
kubectl -n aws-controllers-k8s get subnet
NAME                      SUBNETID                   STATE
subnet-test-private1      subnet-0facxxxxxxxxxxxxx   available
subnet-test-private2      subnet-0ef1xxxxxxxxxxxxx   available
subnet-test-private3
subnet-test-private4
subnet-test-public1       subnet-01eaxxxxxxxxxxxxx   available
subnet-test-public2       subnet-0ae6xxxxxxxxxxxxx   available
subnet-test-rds1          subnet-03a1xxxxxxxxxxxxx   available
subnet-test-rds2          subnet-00f2xxxxxxxxxxxxx   available
```


jaypipes (changes requested):
@smcavallo how about we name the field just "ID" for every resource? Would be shorter and more consistent I think.

@brycahta thoughts?

smcavallo:
> @smcavallo how about we name the field just "ID" for every resource? Would be shorter and more consistent I think.
> 
> @brycahta thoughts?

`ID` would be consistent except it is overloaded since there are so many properties on these objects with IDs.
I would advocate for exposing `ARN` since every object should have one and the name is consistent and unique, except the arns are very long and don't always display well for the table views of kubectl.

brycahta:
> @brycahta thoughts?

+1 to naming the field "ID".

> ID would be consistent except it is overloaded since there are so many properties on these objects with IDs.

That's a fair point; however, I think the format of AWS IDs (prepending ID with resource) will make it clear that the IDs column refers to the resource itself. Ex:
* Vpc: `vpc-<ID>`
* Subnet: `subnet-<ID>`
* Launch Template: `lt-<ID>`

smcavallo:
> > @brycahta thoughts?
> 
> +1 to naming the field "ID".
> 
> > ID would be consistent except it is overloaded since there are so many properties on these objects with IDs.
> 
> That's a fair point; however, I think the format of AWS IDs (prepending ID with resource) will make it clear that the IDs column refers to the resource itself. Ex:
> 
>     * Vpc: `vpc-<ID>`
> 
>     * Subnet: `subnet-<ID>`
> 
>     * Launch Template: `lt-<ID>`

@brycahta  and @jaypipes - I have made the changes and updated the PR description, thanks again!!!

brycahta:
Appreciate the enhancement! @smcavallo

Merging as the test failure is unrelated to your changes (solution in-progress).

brycahta (approved):
/lgtm

brycahta (changes requested):
Missed it on my first pass; can you apply this enhancement to **all resources**?

smcavallo:
@brycahta - I went through and added the outstanding resources which support `ID`
please let me know if I should squash the commits or any other changes are required.

brycahta (changes requested):
* I'm still seeing some resources missing: `DHCPOptions` , `TransitGateway`, `VPCEndpoint`. If there's a reason why we can't include them can you provide some context? Otherwise, please add.
* Did you use [code-generator](https://github.com/aws-controllers-k8s/code-generator) to build the controller? After the updates to `generator.yaml`, I would expect changes to `apis/v1alpha1/ack-generate-metadata.yaml` ([example](https://github.com/aws-controllers-k8s/ec2-controller/pull/101/files#diff-f971dc92ec01edc65e1efcfd75c2f680c939c7e7381ac7ef10589f426382f374R2))
  * Refer to [Building a Controller](https://aws-controllers-k8s.github.io/community/docs/contributor-docs/building-controller/) documentation for more details

smcavallo:
@brycahta - `DHCPOptions` ,` TransitGateway`, `VPCEndpoint` have all been added now
I am using the code-generator but I didn't realize `apis/v1alpha1/ack-generate-metadata.yaml` needed to be checked into source when generating.  I've checked that in as well.

brycahta (changes requested):
Could you remove `-dirty` from the `code-generator` version in `ack-generate-metadata.yaml` please? 

For context, `dirty` is appended automatically if you have any local changes to `code-generator` so if you regen without those changes we should be good to go

smcavallo:
@brycahta - makes sense.  `-dirty` has been removed.

brycahta (approved):
/lgtm

PR #122, opened 2023-01-12
Fix reconciler not detecting differences

RedbackThomson:
Issue #, if available: https://github.com/aws-controllers-k8s/community/issues/1552

Description of changes:
When attempting to debug why conditions were not being set on resources, I discovered that the reconciler was not detecting a difference between the `latest` and `desired` objects' conditions. I determined that this was because `latest.status.conditions` was sharing the same memory as `desired.status.conditions`. Walking through the code with a debugger, I found that it was inside the `Update` code where the memory was being swapped. I have updated these lines in each of the EC2 hook code files so that we no longer return a reference to the `desired` object, and can safely modify the `updated` object independently.


LikithaVemulapalli (approved):
/lgtm 
Thank you for fixing this @RedbackThomson :)

PR #123, opened 2023-01-27
Fix default egress rule deletion logic in case of an egress rule definition

mustafasencer:
**Issue:** https://github.com/aws-controllers-k8s/community/issues/1604

**Description of changes:**
If the user defines an egress rule which is the same as the “default“ egress rule (defined implicitly by AWS), then the ec2-controller raises the below error. Although the SG is created in AWS, the controller is unable to sync the resource.
```txt
2023-01-26T16:19:39.988Z    DEBUG    ackrt    <<<<< rm.syncSGRules    {"account": "647927084307", "role": "", "region": "eu-west-1", "kind": "SecurityGroup", "namespace": "ack-system", "name": "ack-sg-test", "is_adopted": false, "generation": 1}
2023-01-26T16:19:39.988Z    DEBUG    ackrt    <<<< rm.sdkCreate    {"account": "647927084307", "role": "", "region": "eu-west-1", "kind": "SecurityGroup", "namespace": "ack-system", "name": "ack-sg-test", "is_adopted": false, "generation": 1, "error": "InvalidPermission.Duplicate: the specified rule \"peer: 0.0.0.0/0, ALL, ALLOW\" already exists\n\tstatus code: 400, request id: 160a5101-48d5-4de3-afc9-7ff6d7b99f16"}
2023-01-26T16:19:39.988Z    DEBUG    ackrt    <<< rm.Create    {"account": "647927084307", "role": "", "region": "eu-west-1", "kind": "SecurityGroup", "namespace": "ack-system", "name": "ack-sg-test", "is_adopted": false, "generation": 1, "error": "InvalidPermission.Duplicate: the specified rule \"peer: 0.0.0.0/0, ALL, ALLOW\" already exists\n\tstatus code: 400, request id: 160a5101-48d5-4de3-afc9-7ff6d7b99f16"}
2023-01-26T16:19:39.988Z    DEBUG    ackrt    << r.createResource    {"account": "647927084307", "role": "", "region": "eu-west-1", "kind": "SecurityGroup", "namespace": "ack-system", "name": "ack-sg-test", "is_adopted": false, "generation": 1, "error": "InvalidPermission.Duplicate: the specified rule \"peer: 0.0.0.0/0, ALL, ALLOW\" already exists\n\tstatus code: 400, request id: 160a5101-48d5-4de3-afc9-7ff6d7b99f16"}
```
The change in this PR just changes the order of the flow executing the deletion of the “default“ egress rule in case an egress rule is defined by the user. Because the `syncSGRules` during creation does not really sync the SG rules as it does  not have a `latest` state during performing the sync, the “default“ egress rule hangs there and basically causes the issue.

**Note:** I am not sure whether this case requires a specific e2e test case; the current `test_rules_create_update_delete` case covers the situation where only an ingress rule is defined and then it only patches the egress rule definition which does not 100% cover this situation, because this particular case only pops up during the creation step where the default egress rule check occurs. In order to cover it, the egress rule should be defined from the beginning on I suppose. Just wanted to share my limited perspective 😄 

### DIFF
--- a/pkg/resource/security_group/manager.go
+++ b/pkg/resource/security_group/manager.go
@@ -248,6 +248,10 @@ func (rm *resourceManager) LateInitialize(
 func (rm *resourceManager) incompleteLateInitialization(
 	res acktypes.AWSResource,
 ) bool {
+	ko := rm.concreteResource(res).ko.DeepCopy()
+	if ko.Spec.EgressRules == nil {
+		return true
+	}
 	return false
 }
 
--- a/pkg/resource/security_group/manager.go
+++ b/pkg/resource/security_group/manager.go
@@ -51,7 +51,7 @@ var (
 // +kubebuilder:rbac:groups=ec2.services.k8s.aws,resources=securitygroups,verbs=get;list;watch;create;update;patch;delete
 // +kubebuilder:rbac:groups=ec2.services.k8s.aws,resources=securitygroups/status,verbs=get;update;patch
 
-var lateInitializeFieldNames = []string{}
+var lateInitializeFieldNames = []string{"EgressRules"}
 
 // resourceManager is responsible for providing a consistent way to perform
 // CRUD operations in a backend AWS service API for Book custom resources.
@@ -257,7 +257,12 @@ func (rm *resourceManager) lateInitializeFromReadOneOutput(
 	observed acktypes.AWSResource,
 	latest acktypes.AWSResource,
 ) acktypes.AWSResource {
-	return latest
+	observedKo := rm.concreteResource(observed).ko.DeepCopy()
+	latestKo := rm.concreteResource(latest).ko.DeepCopy()
+	if observedKo.Spec.EgressRules != nil && latestKo.Spec.EgressRules == nil {
+		latestKo.Spec.EgressRules = observedKo.Spec.EgressRules
+	}
+	return &resource{latestKo}
 }
 
 // IsSynced returns true if the resource is synced.
--- a/apis/v1alpha1/ack-generate-metadata.yaml
+++ b/apis/v1alpha1/ack-generate-metadata.yaml
@@ -1,5 +1,5 @@
 ack_generate_info:
-  build_date: "2022-10-13T20:54:09Z"
+  build_date: "2023-02-01T15:41:09Z"
   build_hash: 5ee0ac052c54f008dff50f6f5ebb73f2cf3a0bd7
   go_version: go1.18.1
   version: v0.20.1-4-g5ee0ac0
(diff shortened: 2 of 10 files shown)


A-Hilaly:
Thank you @mustafasencer ! QQ: What if the default egress rule was already deleted, should ignore 404 NotFound errors?

jljaco:
Thanks for doing this!  I have created a new e2e test that hopefully will validate this case in 

https://github.com/aws-controllers-k8s/ec2-controller/pull/124

how about we get that one merged and then confirm that the test passes for this change?

jljaco:
> Thanks for doing this! I have created a new e2e test that hopefully will validate this case in
> 
> #124
> 
> how about we get that one merged and then confirm that the test passes for this change?

@mustafasencer #124 is now merged.  In your branch, would you please remove the `@pytest.mark.xfail` line [here](https://github.com/aws-controllers-k8s/ec2-controller/blob/c027a797a7b93a1b3272f8d73097738ca8912153/test/e2e/tests/test_security_group.py#L162)?  That way we can use the new test to confirm that your fix works.

mustafasencer:
@jljaco Thanks for the prompt implementation of the e2e test case. I noticed that the e2e test cases were checking the case where the `spec.egressRules` needed to be updated (due to the “default“ egress rule) after the resource has been synced. Hence, I introduced a late initialisation logic there to account for the `spec.egressRules` field's population after the readOne operation. The e2e test cases are fully covered with this change. Feel free to leave your comments or update the PR directly, as I am not 100% familiar with the codebase, thus have perhaps missed a crucial part.

@A-Hilaly Although the case you mentioned should not be expected, nevertheless I added the ignore logic for the `InvalidPermission.NotFound` in case the default egress rule is not present.

Thanks!

pkg/resource/security_group/manager.go
```diff
@@ -257,7 +257,12 @@ func (rm *resourceManager) lateInitializeFromReadOneOutput(
 	observed acktypes.AWSResource,
 	latest acktypes.AWSResource,
 ) acktypes.AWSResource {
-	return latest
+	observedKo := rm.concreteResource(observed).ko.DeepCopy()
+	latestKo := rm.concreteResource(latest).ko.DeepCopy()
+	if observedKo.Spec.EgressRules != nil && latestKo.Spec.EgressRules == nil {
+		latestKo.Spec.EgressRules = observedKo.Spec.EgressRules
+	}
+	return &resource{latestKo}
```
A-Hilaly, line 265: Can you please include the files you modified to generate this code? `generator.yaml` and what changed in `apis/v1alpha1`

A-Hilaly (commented):
Looks great, thank you @mustafasencer! I have one comment bellow

pkg/resource/security_group/manager.go
mustafasencer: Ohh, sorry I completely forgot about the generation part during manual testing. Pushing the generated code quick...

A-Hilaly (approved):
Neat! Thank you @mustafasencer !

jljaco:
/lgtm

Thank you @mustafasencer !

jljaco:
Thank you @mustafasencer

PR #138, opened 2023-06-06
VPC Flowlogs feature implementation

nnbu:
This change implements flowlogs in ec2 operator.
It adds e2e tests for the flowlogs feature.

Issue #:  https://github.com/aws-controllers-k8s/community/issues/1560

Description of changes:


nnbu:
Please refer my another PR (and the comment) which is created so that code-generator generates correct code for flow-logs.

https://github.com/aws-controllers-k8s/code-generator/pull/452#issuecomment-1579581579

helm/Chart.yaml
```diff
@@ -1,8 +1,8 @@
 apiVersion: v1
 name: ec2-chart
 description: A Helm chart for the ACK service controller for Amazon Elastic Cloud Compute (EC2)
-version: 1.0.3
-appVersion: 1.0.3
+version: 0.0.0-non-release-version
```
jaypipes, line 4: hmm, this is not right :) I think you need to fetch all tags and rebase on main before running `make build-controller`.

Do this:

```
git checkout main && git fetch --all --tags && git rebase upstream/main
```

then git checkout the branch that this PR is in locally:

```
git checkout <YOUR_BRANCH_NAME>
```

and rebase onto main:

```
git rebase main
```

That should bring you up to date with the latest ec2-controller code...

jaypipes: You will also want to rebase your `code-generator` repository onto main after fetching the latest too...

templates/hooks/flow_logs/sdk_create_post_build_request.go.tpl
jaypipes, line 1: Please name this `templates/hooks/flow_log/sdk_create_post_build_request.go.tpl` (singular resource name, not plural :) )

templates/hooks/flow_logs/sdk_file_end.go.tpl
jaypipes, line 1: Please name this `templates/hooks/flow_log/sdk_file_end.go.tpl` (singular resource name, not plural :) )

test/e2e/bootstrap_resources.py
```diff
@@ -15,11 +15,13 @@
 """
 from dataclasses import dataclass
 from acktest.bootstrapping import Resources
+from acktest.bootstrapping.s3 import Bucket
 from acktest.bootstrapping.vpc import VPC
 from e2e import bootstrap_directory
 
 @dataclass
 class BootstrapResources(Resources):
+    FunctionsBucket: Bucket
```
jaypipes, line 24: Can you name this `FlowLogsBucket`? I think you may have copied this from the lambda-controller?

test/e2e/service_bootstrap.py
```diff
@@ -17,14 +17,18 @@
...
 from e2e.bootstrap_resources import BootstrapResources
 
 def service_bootstrap() -> Resources:
     logging.getLogger().setLevel(logging.INFO)
 
     resources = BootstrapResources(
-        SharedTestVPC=VPC(name_prefix="e2e-test-vpc", num_public_subnet=1, num_private_subnet=0)
+        SharedTestVPC=VPC(name_prefix="e2e-test-vpc", num_public_subnet=1, num_private_subnet=0),
+        FunctionsBucket=Bucket(
+            "ack-ec2-controller-tests",
```
jaypipes, line 30: maybe call this `ack-ec2-controller-flow-log-tests`.

Keep in mind bucket names are globally unique :)

jaypipes (changes requested):

generator.yaml
```diff
@@ -7,6 +7,7 @@ ignore:
     - AllocateAddressInput.TagSpecifications
     - CreateDhcpOptionsInput.DryRun
     - CreateDhcpOptionsInput.TagSpecifications
+    - CreateFlowLogsInput.DryRun
```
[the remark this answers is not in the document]
nnbu: done

generator.yaml
```diff
@@ -268,6 +277,28 @@ resources:
...
+        from:
+          operation: CreateTags
+          path: Tags
+        compare:
+          is_ignored: True
+    hooks:
+      delta_pre_compare:
+        code: compareTags(delta, a, b)
+      sdk_create_post_build_request:
+        template_path: hooks/dhcp_options/sdk_create_post_build_request.go.tpl
```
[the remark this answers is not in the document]
nnbu: done

generator.yaml
```diff
@@ -268,6 +277,28 @@ resources:
...
+      sdk_file_end:
+        template_path: hooks/dhcp_options/sdk_file_end.go.tpl
```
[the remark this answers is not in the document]
nnbu: done

templates/hooks/flow_logs/sdk_create_post_build_request.go.tpl
nnbu: done

templates/hooks/flow_logs/sdk_file_end.go.tpl
nnbu: done

test/e2e/service_bootstrap.py
nnbu: done

helm/Chart.yaml
nnbu: I tried these steps. But the generated code always has version as '0.0.0-non-release-version'.

nnbu: RELEASE_VERSION seems to get initialized in https://github.com/aws-controllers-k8s/code-generator/blob/main/scripts/build-controller-release.sh#L159

```
RELEASE_VERSION=${RELEASE_VERSION:-$(git describe --tags --abbrev=0 2>/dev/null || echo $NON_RELEASE_VERSION)}
```

So, I tried executing `git describe --tags` in my code-generator repo. But I get no tags.
```
code-generator % git checkout main && git fetch --all --tags && git rebase origin/main  
Already on 'main'
Your branch is up to date with 'origin/main'.
Current branch main is up to date.
code-generator % git describe --tags
fatal: No names found, cannot describe anything.
```

nnbu: I had missed these steps
```
git remote add upstream git@github.com:aws-controllers-k8s/$REPO
git fetch --all
```

With this done, now release is correctly updated.

apis/v1alpha1/flow_log.go
```diff
@@ -0,0 +1,153 @@
...
+	// resource
+	// +kubebuilder:validation:Optional
+	Conditions []*ackv1alpha1.Condition `json:"conditions"`
+	// Unique, case-sensitive identifier that you provide to ensure the idempotency
+	// of the request.
+	// +kubebuilder:validation:Optional
+	ClientToken *string `json:"clientToken,omitempty"`
+	// The IDs of the flow logs.
+	// +kubebuilder:validation:Optional
+	FlowLogIDs []*string `json:"flowLogIDs,omitempty"`
```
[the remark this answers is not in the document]
jaypipes: You will need to add a tiny bit of custom code in `sdk_read_one_post_set_output` to set the `Status.ID` field to the returned value of `*resp.FlowLogIds[0]` of course, but then you won't need to do things like this in `hooks.go`:

```go
resourceId := []*string{latest.ko.Status.FlowLogIDs[0]}
```

pkg/resource/flow_log/hooks.go
```diff
@@ -0,0 +1,189 @@
...
+import (
+	"context"
+
+	svcapitypes "github.com/aws-controllers-k8s/ec2-controller/apis/v1alpha1"
+	ackcompare "github.com/aws-controllers-k8s/runtime/pkg/compare"
+	ackrtlog "github.com/aws-controllers-k8s/runtime/pkg/runtime/log"
+	svcsdk "github.com/aws/aws-sdk-go/service/ec2"
+)
+
+func (rm *resourceManager) customUpdateFlowLogs(
```
jaypipes, line 25: Recommend renaming this to `customUpdateFlowLog` since we want to make clear that we're treating this "pluralized resource" as a singular resource.

pkg/resource/flow_log/hooks.go
```diff
@@ -0,0 +1,189 @@
...
+	ctx context.Context,
+	desired *resource,
+	latest *resource,
+	delta *ackcompare.Delta,
+) (updated *resource, err error) {
+	rlog := ackrtlog.FromContext(ctx)
+	exit := rlog.Trace("rm.customUpdateFlowLog")
```
jaypipes, line 32: Luckily, you've got the singular already here :)

test/e2e/tests/test_flow_logs.py
```diff
@@ -0,0 +1,149 @@
...
+
+        # Delete k8s resource
+        _, deleted = k8s.delete_custom_resource(ref, 2, 5)
+        assert deleted is True
+
+        time.sleep(DELETE_WAIT_AFTER_SECONDS)
+
+        # Check Flow Log doesn't exist
+        exists = flow_log_exists(ec2_client, resource_id)
+        assert not exists
```
jaypipes, line 128: in a future PR, you may want to add a test case that updates the flow log tags and exercises the `customUpdateFlowLog` function in `hooks.go`

jaypipes (changes requested):

apis/v1alpha1/flow_log.go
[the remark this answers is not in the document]
nnbu: I tried to follow this, but could not get it working.

```
resources:
  FlowLog:
    fields:
      ID:
        type: string
        is_primary_key: true
        is_read_only: true
```
This snippet does not generate a new field ID in apis/v1alpha1/flow_log.go. I must be missing something. I tried removing `is_arn_primary_key: true` as well. But then it fails with error 
```
error calling GoCodeSetResourceIdentifiers: could not find field with path ID for primary key ID
```

Could you guide what I am missing?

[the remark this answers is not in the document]
nnbu: Also, could you tell me more about `sdk_read_one_post_set_output`? I am not sure about how is supposed to work. I think, it is related to `sdkFind`. But I did not understand what code I should add there and where that code will be injected inside `sdkFind` ?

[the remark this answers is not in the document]
nnbu: Somehow, `sdk_read_many_post_set_output` hook is triggered, not `sdk_read_many_post_set_output` . I have following code in the hook which correctly gets injected at the end of `sdkFind`
```
    if found {
        ko.Status.ID = resp.FlowLogs[0].FlowLogId
    }
```

Now, the only issue is `ID` being added. The error is as mentioned in previous comment.
```
error calling GoCodeSetResourceIdentifiers: could not find field with path ID for primary key ID
```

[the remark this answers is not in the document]
nnbu: I understand the dependencies about declaring new `field` in `generator.yaml`.
1. if the field is to go in `spec`, then you can declare using the way you declared (and of course, it cant be `is_read_only` then)
2. if the field is to go in `status` (as we would like), then it needs to have 'from' defined. https://github.com/aws-controllers-k8s/code-generator/blob/main/pkg/model/model.go#L252

So , I defined the field something like this:
```
+.    ID:
+        type: string
+        is_read_only: true
+        from:
+          operation: CreateFlowLogs
+          path: FlowLogIds
```

But for this to work, `FlowLogIds` can't be in 'ignore` section. `#- CreateFlowLogsOutput.FlowLogIds`.
If it is in `ignore` section, then `FlowLogIds` is not `shape.MemberRefs` at https://github.com/aws-controllers-k8s/code-generator/blob/main/pkg/model/sdk_api.go#L396

So, the newly added field `ID` is ignored and not added to `status` struct in generated api.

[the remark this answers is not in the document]
nnbu: Another note here is, all flowlog apis CreateFlowLogs, DescribeFlowLogs, DeleteFlowLogs accept list of flowLogIds as parameter. https://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_DeleteFlowLogs.html

So, even if we convert `FlowLogIds` to `ID` in the `status`, we have to convert that back to list in `newListRequestPayload` and `newDeleteRequestPayload`.

So, I would like to keep `CreateFlowLogsResult.FlowLogIds` as it is. Let me know your thoughts.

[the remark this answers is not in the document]
jaypipes: I don't see the point of keeping it a list when there will only ever be a single value in there. :)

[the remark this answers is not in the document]
jaypipes: > Also, could you tell me more about `sdk_read_one_post_set_output`? I am not sure about how is supposed to work. I think, it is related to `sdkFind`. But I did not understand what code I should add there and where that code will be injected inside `sdkFind` ?

Yes, for each of the CRUD actions (`read_[one|many]`, `create`, `update` and `delete`) there is a corresponding custom code hook point called `sdk_$action_post_set_output` that is executed immediately after the response Output shape is processed and the custom resource's fields are set to values in that response output.

If we take a look at the `DHCPOptions` resource's `sdkFind` method, this line here:

https://github.com/aws-controllers-k8s/ec2-controller/blob/5d460dbad24e44a11e80a693fbd61d498820476b/pkg/resource/dhcp_options/sdk.go#L147

is where the `sdk_read_many_post_set_output` hook would be executed if there was one. It is an opportunity to override any fields on the [`ko`](https://github.com/aws-controllers-k8s/ec2-controller/blob/5d460dbad24e44a11e80a693fbd61d498820476b/pkg/resource/dhcp_options/sdk.go#L87) variable that represents the custom resource being returned from the function.

So, in the case of FlowLog, we could put the following code into the `sdk_read_many_post_set_output` custom code hook:

```go
ko.Status.ID = *resp.FlowLogs[0].ResourceId
```

because the [`DescribeFlowLogs`](https://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_DescribeFlowLogs.html) response uses the field name `ResourceId` instead of `FlowLogId` ( :( for consistency...).

In the `sdk_create_post_set_output` code hook, you'd do this:

```go
ko.Status.ID = *resp.FlowLogIds[0]
```

because the `CreateFlowLogs` response field is called `FlowLogIds`.

pkg/resource/flow_log/hooks.go
nnbu: done

test/e2e/tests/test_flow_logs.py
nnbu: sure

apis/v1alpha1/flow_log.go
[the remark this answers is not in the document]
nnbu: > I don't see the point of keeping it a list when there will only ever be a single value in there. :)

After reviewing the CreateFlowLogs, DescribeFlowLogs, DeleteFlowLogs, we are working with lists always. e.g. CreateFlowLogs allows to create a list of of FlowLogsIds. 

So, I modified e2e tests as well to accommodate creation of multiple FlowLogIds. The e2e tests are passing.

I also changed this code 
```
resourceId := []*string{latest.ko.Status.FlowLogIDs[0]}
```
to
```
resourceId := latest.ko.Status.FlowLogIDs
```

So, I am no longer checking single FlowLogId.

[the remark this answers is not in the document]
nnbu: As discussed over the chat, I have now converted FlowLogIds, ResourceIDs (list) to FlowLogId, ResourceID (single string)

generator.yaml
```diff
@@ -268,6 +281,42 @@ resources:
...
+      - InvalidParameter
+      - InvalidParameterValue
+    fields:
+      ResourceID:
+        type: string
+        is_required: true
+      ID:
+        type: string
+        is_read_only: true
+        is_primary_key: true
```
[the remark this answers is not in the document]
nnbu: In fact it worked after this change in code-generator.
https://github.com/aws-controllers-k8s/code-generator/pull/453

apis/v1alpha1/ack-generate-metadata.yaml
```diff
@@ -1,13 +1,13 @@
 ack_generate_info:
-  build_date: "2023-05-15T23:06:36Z"
-  build_hash: 8f3ba427974fd6e769926778d54834eaee3b81a3
-  go_version: go1.19
-  version: v0.26.1
-api_directory_checksum: 6e86aa4f26729ce014485b1096286db654459af7
+  build_date: "2023-06-09T22:43:08Z"
+  build_hash: b74f41dae816d0038170066823283cfc0f981bbd
+  go_version: go1.20.3
+  version: v0.26.1-5-gb74f41d-dirty
```
LikithaVemulapalli, line 5: This version shouldn't be tagged with dirty, could you try fetching the tags `git fetch --tags` in `code-generator` and regenerate the code. Thanks :)

LikithaVemulapalli (changes requested):

apis/v1alpha1/ack-generate-metadata.yaml
nnbu: I missed it. I needed https://github.com/aws-controllers-k8s/code-generator/pull/453 to be merged before regenerating the code. Now, it is merged. and I have uploaded the change after new code generation.

LikithaVemulapalli (approved):
Great work @nnbu :)

PR #139, opened 2023-06-07
fixing dropping of secondary cidr blocks on first apply of VPC

nishant221:
Issue https://github.com/aws-controllers-k8s/community/issues/1816

Description of changes:
Added the fix for a bug where Secondary VPC CIDR blocks are getting dropped in the first apply of VPC resource.

### DIFF
--- a/apis/v1alpha1/ack-generate-metadata.yaml
+++ b/apis/v1alpha1/ack-generate-metadata.yaml
@@ -1,9 +1,9 @@
 ack_generate_info:
-  build_date: "2023-06-14T17:05:27Z"
-  build_hash: e04fca8e1fa32e988ff4dd9425d758ecc8a2a178
-  go_version: go1.20.3
-  version: v0.26.1-6-ge04fca8
-api_directory_checksum: a930caa16911411d04c5e7c14b27e077993d93a1
+  build_date: "2023-06-17T15:45:55Z"
+  build_hash: e9b68590da73ce9143ba1e4361cebdc1d876c81e
+  go_version: go1.19
+  version: v0.26.1-7-ge9b6859
+api_directory_checksum: 9452be5c34acba71fcd740b2acf4d053f2eb221b
 api_version: v1alpha1
 aws_sdk_go_version: v1.44.93
 generator_config_info:
--- a/pkg/resource/vpc/hooks.go
+++ b/pkg/resource/vpc/hooks.go
@@ -18,7 +18,6 @@ import (
 
 	svcapitypes "github.com/aws-controllers-k8s/ec2-controller/apis/v1alpha1"
 	ackcompare "github.com/aws-controllers-k8s/runtime/pkg/compare"
-	ackerr "github.com/aws-controllers-k8s/runtime/pkg/errors"
 	ackrtlog "github.com/aws-controllers-k8s/runtime/pkg/runtime/log"
 	svcsdk "github.com/aws/aws-sdk-go/service/ec2"
 )
@@ -173,9 +172,11 @@ func (rm *resourceManager) syncCIDRBlocks(
 	latestCIDRs := latest.ko.Spec.CIDRBlocks
 	latestCIDRStates := latest.ko.Status.CIDRBlockAssociationSet
 	toAddCIDRs, toDeleteCIDRs := computeStringPDifference(desiredCIDRs, latestCIDRs)
+	cidrblockassociationset := []*svcapitypes.VPCCIDRBlockAssociation{}
 
 	// extract associationID for the DisassociateVpcCidr request
 	for _, cidr := range toDeleteCIDRs {
+
 		input := &svcsdk.DisassociateVpcCidrBlockInput{}
 		for _, cidrAssociation := range latestCIDRStates {
 			if *cidr == *cidrAssociation.CIDRBlock {
@@ -185,6 +186,8 @@ func (rm *resourceManager) syncCIDRBlocks(
 				if err != nil {
 					return err
 				}
+			} else {
+				cidrblockassociationset = append(cidrblockassociationset, cidrAssociation)
 			}
 		}
 	}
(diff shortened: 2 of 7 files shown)


LikithaVemulapalli:
Hello @nishant221, we don't generally add code in `sdk.go` manually, you need to re-generate the controller. We follow the pattern of adding [templates/hooks](https://github.com/aws-controllers-k8s/ec2-controller/tree/main/templates/hooks/vpc ) to include code in `sdk`. You need to figure out if these changes needs to be added in the end or after create, and include them in the respective file. In general we will be calling these files in [`generator.yaml`](https://github.com/aws-controllers-k8s/ec2-controller/blob/5d460dbad24e44a11e80a693fbd61d498820476b/generator.yaml#L592) and when we run `make build-controller SERVICE=ec2` from code-generator, changes will be included in sdk file.

nishant221:
@LikithaVemulapalli  Thanks for the feedback/review . Will make the required changes and update the PR.

nishant221:
Hi @LikithaVemulapalli ,

As per the suggestion, tried making the required changes. Kindly have a look.
Thanks

apis/v1alpha1/ack-generate-metadata.yaml
```diff
@@ -1,8 +1,8 @@
 ack_generate_info:
-  build_date: "2023-05-15T23:06:36Z"
-  build_hash: 8f3ba427974fd6e769926778d54834eaee3b81a3
+  build_date: "2023-06-08T05:23:27Z"
+  build_hash: 6542e04cc357215c91df94e747e3d412fb26df97
   go_version: go1.19
-  version: v0.26.1
+  version: 6542e04-dirty
```
LikithaVemulapalli, line 5: This version seems incorrect! There shouldn't be any dirty tagged along with the version, could you please revert your `GOPROXY` changes if made and try regenerating the code.

LikithaVemulapalli (commented):
Could you also include a simple e2e test to validate this scenario, Thanks :)

apis/v1alpha1/ack-generate-metadata.yaml
RedbackThomson: I think the issue here might also be that you need to fetch the tags. Try `git fetch --tags` in the `code-generator` repository

RedbackThomson (commented):
This is excellent! Thank you for taking the time to work on this.

There are still 2 cases that I think we might need to support:
- What happens when a user updates the list of CIDR blocks? Do we need to remove the old associations and add the new ones?
- What happens when we delete the VPC altogether? Do we need to take any actions to delete the CIDR blocks before doing anything else?

apis/v1alpha1/ack-generate-metadata.yaml
nishant221: Regenerated the code. Now the `version` is not getting tagged with `-dirty` .

@RedbackThomson  Thanks for the suggestion. It was fixed by doing both `git fetch --all --tags` and `git update-index --refresh`.

nishant221:
@RedbackThomson  Thanks for the suggestions/review .


> What happens when a user updates the list of CIDR blocks? Do we need to remove the old associations and add the new ones?

As described in the issue description (https://github.com/aws-controllers-k8s/community/issues/1816 ) , the functionality (or updates) are working fine with object reapply. Hence, I believe no changes are needed with updates.
 
> What happens when we delete the VPC altogether? Do we need to take any actions to delete the CIDR blocks before doing anything else?
Deleting the VPC does not require any additional action to be taken.

LikithaVemulapalli:
discussed offline, adding `ok-to-test` label to see if the tests pass

RedbackThomson:
I have a hunch that if someone removes this secondary CIDR block from the list, that it won't be properly disassociated. That is because we are never calling [`DisassociateVpcCidrBlock`](https://docs.aws.amazon.com/sdk-for-go/api/service/ec2/#EC2.DisassociateVpcCidrBlock) in the hook code update logic. 

Would you mind adding a test for this condition (after creating the resource, remove a CIDR block and check the resource with boto3)?

nishant221:
@RedbackThomson I think [DisassociateVpcCidrBlock](https://docs.aws.amazon.com/sdk-for-go/api/service/ec2/#EC2.DisassociateVpcCidrBlock) is getting called [here](https://github.com/aws-controllers-k8s/ec2-controller/blob/main/pkg/resource/vpc/hooks.go#L182) via [syncCIDRBlocks](https://github.com/aws-controllers-k8s/ec2-controller/blob/main/pkg/resource/vpc/hooks.go#L256) .

The reason why it's not updating the AWS VPC resource is a small bug in  [computeStringPDifference function] (https://github.com/aws-controllers-k8s/ec2-controller/pull/139/files#diff-8e9c68bc9772730bc1c3837f6adeefebab8a5752e52088e9a7138ae123a52f3cR152). Added the fix the PR.

Also, added an additional test as suggested. E2E test seems to be passing. Kindly have a look.

Thanks

pkg/resource/vpc/hooks.go
```diff
@@ -258,6 +259,50 @@ func (rm *resourceManager) customUpdateVPC(
 		}
 	}
 
+	// CidrBlockAssociationSet needs to be updated in the Status
```
RedbackThomson, line 262: See my comments below - but these new lines shouldn't be necessary if we re-use `syncCIDRBlocks`

pkg/resource/vpc/hooks.go
```diff
@@ -428,3 +473,48 @@ func updateTagSpecificationsInCreateRequest(r *resource,
 		input.TagSpecifications = []*svcsdk.TagSpecification{&desiredTagSpecs}
 	}
 }
+
+func (rm *resourceManager) attachSecondaryCidrBlocks(ctx context.Context, desired *resource,
```
RedbackThomson, line 477: This code is essentially duplicating the logic already inside `syncCIDRBlocks` - except it uses a `CreateVpcOutput` rather than the resource status. I believe `sdk_create_post_set_output` hook code can call `syncCIDRBlocks` directly. We do the same thing for [S3 `Bucket`](https://github.com/aws-controllers-k8s/s3-controller/blob/main/pkg/resource/bucket/hook.go#L57)

I do see that you are using the response from this SDK call to set the `CidrBlockAssociation` struct, though, and I think this is preferable to calling `DescribeVpcs` again (since it doesn't require additional API calls). We don't have this logic in the body of `syncCIDRBlocks`, so it would be great if you could move that into there.

RedbackThomson (changes requested):

pkg/resource/vpc/hooks.go
nishant221: @RedbackThomson Updated the PR with the recommended changes.

RedbackThomson: I don't see a call to `syncCIDRBlocks` in any code paths coming off create. Could you please add logic within `sdk_create_post_set_output`?

nishant221: I have already added the logic within `sdk_create_post_set_output` 
https://github.com/aws-controllers-k8s/ec2-controller/pull/139/files#diff-533c8b725d5ee9f8fc14e5e3f070d31296017f0f5c6cd966a319219076948c97R5

pkg/resource/vpc/hooks.go
nishant221: modification done as suggested.

pkg/resource/vpc/hooks.go
RedbackThomson: Oh my bad, when I clicked around I didn't see it. Must have missed it

RedbackThomson (approved):

PR #142, opened 2023-06-21
Fix for Circular dependency while creating RouteTable and a VPCEndpoint

nishant221:
Issue: https://github.com/aws-controllers-k8s/community/issues/1817

Description of changes:
This is for fixing circular dependency bug while creating a new RouteTable and a VPCEndpoint. As per the changes in this PR, any route to `DestinationPrefixList` which is owned by "AWS" will be ignored while syncing routes and will be used as is.

Tested via following :

```
kind: VPC
metadata:
  name: circular-dependency-vpc
spec:
  cidrBlocks:
  - 10.10.0.0/16
  enableDNSSupport: True
  enableDNSHostnames: True
  tags:
    - key: Name
      value: circular-dependency-vpc
---
kind: RouteTable
metadata:
  name: circular-dependency-rtb
spec:
  routes:
    - destinationCIDRBlock: 0.0.0.0/0
      gatewayRef:
        from:
          name: circular-dependency-igw
  vpcRef:
    from:
      name: circular-dependency-vpc
---
kind: VPCEndpoint
metadata:
  name:  circular-dependency-vpc-endpoint-s3
spec:
  serviceName: com.amazonaws.us-west-2.s3
  vpcRef:
    from:
      name: circular-dependency-vpc
  routeTableRefs:
  - from:
      name: circular-dependency-rtb
  tags:
    - key: Name
      value:  circular-dependency-vpc-endpoint-s3
---
kind: InternetGateway
metadata:
  name: circular-dependency-igw
spec:
  vpcRef:
    from:
      name: circular-dependency-vpc
  tags:
    - key: "Adobe.Owner"
      value: "Ethos"
    - key: Name
      value: circular-dependency-igw
```

Status (before reconciliation i.e. before ACK synced from AWS actual resources):

```
$ kubectl get routetable circular-dependency-rtb -o json | jq .status.routeStatuses
[
  {
    "destinationCIDRBlock": "10.10.0.0/16",
    "gatewayID": "local",
    "origin": "CreateRouteTable",
    "state": "active"
  },
  {
    "destinationCIDRBlock": "0.0.0.0/0",
    "gatewayID": "igw-06d5825a3e9ba9dee",
    "origin": "CreateRoute",
    "state": "active"
  }
]
```

Status (after reconciliation i.e. after ACK synced from AWS actual resources):

```
[
  {
    "destinationCIDRBlock": "10.10.0.0/16",
    "gatewayID": "local",
    "origin": "CreateRouteTable",
    "state": "active"
  },
  {
    "destinationCIDRBlock": "0.0.0.0/0",
    "gatewayID": "igw-06d5825a3e9ba9dee",
    "origin": "CreateRoute",
    "state": "active"
  },
  {
    "destinationPrefixListID": "pl-68a54001",
    "gatewayID": "vpce-017f902c0d2c007e3",
    "origin": "CreateRoute",
    "state": "active"
  }
]
```

### DIFF
--- a/pkg/resource/route_table/hooks.go
+++ b/pkg/resource/route_table/hooks.go
@@ -21,6 +21,7 @@ import (
 	ackcompare "github.com/aws-controllers-k8s/runtime/pkg/compare"
 	ackrtlog "github.com/aws-controllers-k8s/runtime/pkg/runtime/log"
 	svcsdk "github.com/aws/aws-sdk-go/service/ec2"
+	"github.com/samber/lo"
 )
 
 const LocalRouteGateway = "local"
@@ -307,33 +308,41 @@ func (rm *resourceManager) excludeAwsRoute(
 	// Preparing a list of prefixIds from all the DestinationPrefixListIDs in Routes
 	// This is to prevent multiple AWS API calls of DescribeManagedPrefixLists
 
-	for _, route := range routes {
-		if route.DestinationPrefixListID != nil {
-			prefixListIds = append(prefixListIds, route.DestinationPrefixListID)
-		}
-	}
+	ret = lo.Reject(routes, func(route *svcapitypes.CreateRouteInput, index int) bool {
+		return route.DestinationPrefixListID != nil
+	})
+
+	prefix_list_routes := lo.Filter(routes, func(route *svcapitypes.CreateRouteInput, index int) bool {
+		return route.DestinationPrefixListID != nil
+	})
+
+	prefixListIds = lo.Map(prefix_list_routes, func(route *svcapitypes.CreateRouteInput, _ int) *string {
+		return route.DestinationPrefixListID
+
+	})
 
 	var resp *svcsdk.DescribeManagedPrefixListsOutput
-	awsMpl := map[string]struct{}{}
 	input := &svcsdk.DescribeManagedPrefixListsInput{}
 	input.PrefixListIds = prefixListIds
 	resp, _ = rm.sdkapi.DescribeManagedPrefixLists(input)
 	rm.metrics.RecordAPICall("READ_MANY", "DescribeManagedPrefixLists", nil)
 
-	for _, mpl := range resp.PrefixLists {
+	m := lo.FilterMap(resp.PrefixLists, func(mpl *svcsdk.ManagedPrefixList, _ int) (string, bool) {
 		if strings.EqualFold(*mpl.OwnerId, "AWS") {
-			awsMpl[*mpl.PrefixListId] = struct{}{}
+			return *mpl.PrefixListId, true
 		}
-	}
-	for _, route := range routes {
-		if route.DestinationPrefixListID != nil {
-			if _, found := awsMpl[*route.DestinationPrefixListID]; !found {
-				ret = append(ret, route)
-			}
-		} else {
-			ret = append(ret, route)
+		return "", false
+	})
+
+	filtered_routes := lo.FilterMap(prefix_list_routes, func(route *svcapitypes.CreateRouteInput, _ int) (*svcapitypes.CreateRouteInput, bool) {
+		found := lo.IndexOf(m, *route.DestinationPrefixListID)
+		if found == -1 {
+			return route, true
 		}
-	}
+		return nil, false
+	})
+	ret = append(ret, filtered_routes...)
+
 	return ret
 }
 
--- a/pkg/resource/route_table/hooks.go
+++ b/pkg/resource/route_table/hooks.go
@@ -45,8 +45,13 @@ func (rm *resourceManager) syncRoutes(
 	defer exit(err)
 	toAdd := []*svcapitypes.CreateRouteInput{}
 	toDelete := []*svcapitypes.CreateRouteInput{}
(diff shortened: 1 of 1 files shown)


pkg/resource/route_table/hooks.go
```diff
@@ -290,6 +298,45 @@ func removeLocalRoute(
...
+	}
+	for _, route := range routes {
+		if route.DestinationPrefixListID != nil {
+			if _, found := awsMpl[*route.DestinationPrefixListID]; !found {
+				ret = append(ret, route)
+			}
+		} else {
+			ret = append(ret, route)
+		}
+	}
```
RedbackThomson, line 336: nit: Is there some way we could make this simpler using the `lo.Filter` utility? https://github.com/samber/lo#filter

pkg/resource/route_table/hooks.go
```diff
@@ -290,6 +298,45 @@ func removeLocalRoute(
...
+	var prefixListIds []*string
+
+	// Preparing a list of prefixIds from all the DestinationPrefixListIDs in Routes
+	// This is to prevent multiple AWS API calls of DescribeManagedPrefixLists
+
+	for _, route := range routes {
+		if route.DestinationPrefixListID != nil {
+			prefixListIds = append(prefixListIds, route.DestinationPrefixListID)
+		}
+	}
```
RedbackThomson, line 314: nit: This can be replaced with a `lo.Map` - https://github.com/samber/lo#map

nishant221: @RedbackThomson Thanks for the review. This can be replaced with following but I think this will add complexity and with the use of additional library. Actually we have to create a `prefixListIds` first to club `DescribeManagedPrefixLists` AWS call into a single. Kindly suggest.



```
prefixListIds = lo.FilterMap(routes, func(route *svcapitypes.CreateRouteInput, _ int) (*string, bool) {
		if route.DestinationPrefixListID != nil {
			return route.DestinationPrefixListID, true
		}
		return nil, false
	})
```

pkg/resource/route_table/hooks.go
nishant221: Although it can be done (sample code below, haven't tested the below one though). As mentioned in other comment, we maybe adding complexity to the code if we use a new library here. Kindly have a look and suggest.

```
       ret = lo.Reject(routes, func(route *svcapitypes.CreateRouteInput, index int) bool {
		return route.DestinationPrefixListID != nil
	})

	prefix_list_routes := lo.Filter(routes, func(route *svcapitypes.CreateRouteInput, index int) bool {
		return route.DestinationPrefixListID != nil
	})

	prefixListIds = lo.Map(prefix_list_routes, func(route *svcapitypes.CreateRouteInput, _ int) *string {
		return route.DestinationPrefixListID

	})


	var resp *svcsdk.DescribeManagedPrefixListsOutput
	input := &svcsdk.DescribeManagedPrefixListsInput{}
	input.PrefixListIds = prefixListIds
	resp, _ = rm.sdkapi.DescribeManagedPrefixLists(input)
	rm.metrics.RecordAPICall("READ_MANY", "DescribeManagedPrefixLists", nil)


        m := lo.FilterMap(resp.PrefixLists, func(mpl *svcsdk.ManagedPrefixList, _ int) (string, bool) {
		if strings.EqualFold(*mpl.OwnerId, "AWS") {
			return *mpl.PrefixListId, true
		}
		return "", false
	})


        filtered_routes := lo.FilterMap(prefix_list_routes, func(route *svcapitypes.CreateRouteInput, _ int) (*svcapitypes.CreateRouteInput, bool) {
		found := lo.IndexOf(m, *route.DestinationPrefixListID)
		if found != -1 {
			return route, true
		}
		return nil, false
	})
	ret = append(ret, filtered_routes...)

	return ret


```

nishant221: As recommended, updated the code using "Io". Kindly have a look.
Thanks

pkg/resource/route_table/hooks.go
nishant221: As recommended, updated the code using "Io". Kindly have a look.
Thanks

nishant221:
@RedbackThomson As recommended, updated the code via "Io". kindly have a look.
Thanks

pkg/resource/route_table/hooks.go
```diff
@@ -290,6 +299,53 @@ func removeLocalRoute(
...
+	m := lo.FilterMap(resp.PrefixLists, func(mpl *svcsdk.ManagedPrefixList, _ int) (string, bool) {
+		if strings.EqualFold(*mpl.OwnerId, "AWS") {
+			return *mpl.PrefixListId, true
+		}
+		return "", false
+	})
+
+	filtered_routes := lo.FilterMap(prefix_list_routes, func(route *svcapitypes.CreateRouteInput, _ int) (*svcapitypes.CreateRouteInput, bool) {
+		found := lo.IndexOf(m, *route.DestinationPrefixListID)
+		if found == -1 {
```
RedbackThomson, line 339: super nit: We could use `lo.Contains` here. This does work, too.

RedbackThomson (approved):

PR #149, opened 2023-09-13
Add route table assocations to internet gateway

nnbu:
Issue # https://github.com/aws-controllers-k8s/community/issues/1892

Description of changes:
New custom field RouteTables is added to internetGateway CRD. Route Table to Internet Gateway association is done using AssociateRouteTable API call and association is removed using DisassociateRouteTable API call.


nnbu:
I have followed the same model that introduced [routeTables](https://github.com/aws-controllers-k8s/ec2-controller/blob/685cd7e0fec41dc0d7c9563f3e1c8e0a69a4c52d/apis/v1alpha1/subnet.go#L67) field in subnet spec.

a-hilaly (commented):
hold as we will release a new code gen version in the next few hours/days.
/ok-to-test
/hold

nnbu:
> hold as we will release a new code gen version in the next few hours/days. /ok-to-test /hold

Were you referencing this issue https://github.com/aws-controllers-k8s/community/issues/1893 ? Now that it is closed, do I need to generate the code again?

a-hilaly:
@nnbu yes, there is some rebase conflicts, ideally you should use `code-generator` 0.27.1, thank you!

nnbu:
@RedbackThomson Will you be reviewing this PR please? It is similar to the change you did for subnet to routeTable association.

pkg/resource/internet_gateway/hooks.go
```diff
@@ -281,3 +288,192 @@ func updateTagSpecificationsInCreateRequest(r *resource,
...
+		exit(err)
+	}(err)
+
+	existingRTs, err := rm.getRouteTableAssociations(ctx, latest)
+	if err != nil {
+		return err
+	}
+
+	toAdd := make([]*string, 0)
+	toDelete := make([]*svcsdk.RouteTableAssociation, 0)
```
a-hilaly, line 333: just for consistency can we make this array of string type? `make([]*string, 0)`

pkg/resource/internet_gateway/hooks.go
```diff
@@ -281,3 +288,192 @@ func updateTagSpecificationsInCreateRequest(r *resource,
...
+		exit(err)
+	}(err)
+
+	input := &svcsdk.DescribeRouteTablesInput{}
+
+	for {
+		resp, err := rm.sdkapi.DescribeRouteTablesWithContext(ctx, input)
+		if err != nil || resp == nil {
+			break
+		}
```
a-hilaly, line 436: I believe we should just return an error here. The break when `resp.NextToken == nil|""`  is enough

pkg/resource/internet_gateway/hooks.go
```diff
@@ -281,3 +288,192 @@ func updateTagSpecificationsInCreateRequest(r *resource,
...
+
+		rm.metrics.RecordAPICall("GET", "DescribeRouteTables", err)
+		for _, rt := range resp.RouteTables {
+			var assoc *svcsdk.RouteTableAssociation
```
a-hilaly, line 440:
```suggestion
			var rtAssociations *svcsdk.RouteTableAssociation
```

pkg/resource/internet_gateway/hooks.go
```diff
@@ -281,3 +288,192 @@ func updateTagSpecificationsInCreateRequest(r *resource,
...
+			// Find the association for the current internet gateway
+			for _, as := range rt.Associations {
+				if as.GatewayId == nil || res.ko.Status.InternetGatewayID == nil {
+					continue
+				}
+				if *as.GatewayId == *res.ko.Status.InternetGatewayID && *as.AssociationState.State == "associated" {
+					assoc = as
+					break
+				}
```
a-hilaly, line 449: Maybe just do the append here?

a-hilaly (commented):
Thank you @nnbu ! Left few comments in-line

pkg/resource/internet_gateway/hooks.go
nnbu: I could do that. However, I am using `svcsdk.RouteTableAssociation` to print Debug message with 2 fields 1. route table id and 2. association id. `*[]string` in case of toDelete would need to hold association id only.

In case of toAdd, `association` is not established yet. so, there is no association id to print.

I can still make the change and remove the debug statement, because printing association id without mentioning routing table id would be not much of a use.

pkg/resource/internet_gateway/hooks.go
nnbu: I am not fully sure, but if resp is nil ever (without err being nil), then the for loop at line 439 would crash while dereferencing `resp`.

test/e2e/tests/test_internet_gateway.py
```diff
@@ -249,4 +284,35 @@ def test_crud_tags(self, ec2_client, simple_internet_gateway):
         time.sleep(DELETE_WAIT_AFTER_SECONDS)
 
         # Check IGW no longer exists in AWS
-        ec2_validator.assert_internet_gateway(resource_id, exists=False)
\ No newline at end of file
+        ec2_validator.assert_internet_gateway(resource_id, exists=False)
+
+def simple_route_table(vpc_id: int):
```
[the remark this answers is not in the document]
nnbu: I did think about before copying over this function here. :) 
But this one is `simpler` route table that does not have additional entries associated with like. e.g. IGW_ID(internet gateway), DEST_CIDR_BLOCK  among others. 

https://github.com/aws-controllers-k8s/ec2-controller/blob/main/test/e2e/tests/test_route_table.py#L47C5-L49C1

It would need unnecessary functionality to create additional internet gateway while this test is testing internet gateway itself.

pkg/resource/internet_gateway/hooks.go
nnbu: done

pkg/resource/internet_gateway/hooks.go
nnbu: done

pkg/resource/internet_gateway/hooks.go
a-hilaly: I see what you did there. Good points.

test/e2e/tests/test_internet_gateway.py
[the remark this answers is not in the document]
a-hilaly: I see that you're having a separate resource file for this function, which make things ok! Great!

a-hilaly (approved):
Thank you very much sir! @nnbu 
/lgtm

PR #154, opened 2023-10-17
Ec2: Network ACL CR should support deletion when subnet association is present

nnbu:
Issue #, if available:

Description of changes:
Currently, when network acl cr has subnet association configured, we can not delete the cr. subnet association needs to be removed first and then cr can be deleted. This PR addresses this issue. When network acl CR is being deleted, the new change deletes subnet association from aws first and then deletes network acl from aws. This allows CR deletion to happen without any problem.


nnbu:
@a-hilaly Please review when you get a chance

pkg/resource/network_acl/hooks.go
```diff
@@ -194,6 +194,9 @@ func (rm *resourceManager) syncAssociation(
 	exit := rlog.Trace("rm.syncAssociation")
 	defer exit(err)
 
+	if desired == nil && latest == nil {
+		return nil
+	}
```
a-hilaly, line 199: Correct me if i'm wrong, but i don't think this is needed (or at least need some subset of it). If we reach `createAssociation` it means that we've reached `sdkCreate`, which means 100% that desired is not nil, or `sdkUpdate` which means that both of desired and latest are not nil.

pkg/resource/network_acl/hooks.go
```diff
@@ -245,6 +248,16 @@ func (rm *resourceManager) deleteOldAssociations(
...
+	var vpcID *string
+	var aclID *string
+	if desired != nil {
+		vpcID = desired.ko.Spec.VPCID
+		aclID = desired.ko.Status.ID
+	} else {
+		vpcID = latest.ko.Spec.VPCID
+		aclID = latest.ko.Status.ID
+
+	}
```
a-hilaly, line 260: Same philosophy comes back here, if we're reaching this level, it means that the resource exists, meaning that desired != nil and latest != nil

nnbu: `deleteOldAssociations` is called from `syncAssociation`, which in turn is called from either of `sdkCreate`, `sdkUpdate` or `sdkDelete`.

In case of `sdkDelete`, `desired` will be nil. 
In case of `sdkCreate`, `latest` will be nil.

pkg/resource/network_acl/hooks.go
nnbu: `syncAssociation` is called in 3 cases. `sdkCreate`, `sdkUpdate` and `sdkDelete`. I initially thought this would be needed for `sdkDelete` case. But you are right. This is not needed and I will remove this block.

nnbu:
I have addressed the comments. Please review when you get a chance. @a-hilaly

a-hilaly (approved):
Thank you @nnbu !
/lgtm

PR #156, opened 2023-10-26
Fix critical error in flowlogs pkg

mattzech:
Issue #, if available: [1931](https://github.com/aws-controllers-k8s/community/issues/1931)

Description of changes:
https://pkg.go.dev/github.com/aws/aws-sdk-go-v2/service/ec2#DescribeFlowLogsOutput

FlowLogID is not a field defined in DescribeFlowlogsOutput. Need to go inside the resp.FlowLog[] to get the FlowLogID. This is causing a crash any time FlowLog resource is created because it is always going to return index out of bounds.

### DIFF
--- a/pkg/resource/flow_log/sdk.go
+++ b/pkg/resource/flow_log/sdk.go
@@ -248,8 +248,8 @@ func (rm *resourceManager) sdkCreate(
 	}
 
 	rm.setStatusDefaults(ko)
-	if resp.FlowLogIds[0] != nil {
-		ko.Status.FlowLogID = resp.FlowLogIds[0]
+	if resp.FlowLogs[0] != nil {
+		ko.Status.FlowLogID = resp.FlowLogs[0].FlowLogID
 	}
 	return &resource{ko}, nil
 }


mattzech:
I realize my code may not be exactly right, I am just basing this off my knowledge of the sdk.

a-hilaly (commented):
Hi @mattzech - thanks for the PR! I just checked [`CreateFlowLogsOutput`](https://docs.aws.amazon.com/sdk-for-go/api/service/ec2/#CreateFlowLogsOutput) struct - and it looks like it doesn't have a `FlowLogs` field.

a-hilaly:
@mattzech I just realized that you are pointing to an sdk v2 link, while in ACK we are using the v1

mattzech:
@a-hilaly Ah, that would make sense as to why I did not see it. Thank you for the response. Did you get a chance to look at the issue I referenced? Any basic FlowLogs resource creation panics and points to this line in the code

a-hilaly:
Taking a look now!

a-hilaly:
Hi @mattzech - sorry for the delays. the since the sdk files are generated we  needed a hook to be used here. It happened that @nnbu just fixed that in #157  - so we can close this one. 

Thank you both!!

PR #157, opened 2023-10-26
Fix flowlog crash

nnbu:
Issue #, if available: [#1931](https://github.com/aws-controllers-k8s/community/issues/1931)

Description of changes:
While creating FlowLog, if logDestnation does not exist then CreateFlowLogsWithContext does not return the error. However, Unsuccessful field in the response is set. FlowLog is not created in aws. Due to this, controller crashes while accessing the flow log id from the response (resp.FlowLogIds[0]).
Fixing the issue by checking if FlowLogIds has valid length before accessing it.

### DIFF
--- a/apis/v1alpha1/ack-generate-metadata.yaml
+++ b/apis/v1alpha1/ack-generate-metadata.yaml
@@ -1,5 +1,5 @@
 ack_generate_info:
-  build_date: "2024-09-17T19:14:20Z"
+  build_date: "2024-09-20T17:31:34Z"
   build_hash: f8f98563404066ac3340db0a049d2e530e5c51cc
   go_version: go1.22.6
   version: v0.38.1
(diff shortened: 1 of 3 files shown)


config/controller/kustomization.yaml
```diff
@@ -6,4 +6,4 @@ kind: Kustomization
 images:
 - name: controller
   newName: public.ecr.aws/aws-controllers-k8s/ec2-controller
-  newTag: 1.0.7
+  newTag: 1.0.8
```
a-hilaly, line 9: Can we at least split the release and the feature into two separate commits? think about making reverts easier if ever needed

nnbu: Done

a-hilaly (approved):
/lgtm

PR #158, opened 2023-11-23
feat(VPCPeeringConnection): add support for custom resource and additional fields for Modify OP and auto-accepting

marcdavoli:
## Changes
This PR adds the `vpcpeeringconnections` custom resource via the standard code-generator output, but with a few custom additions:
- Adds the possibility to use an ACK VPC resource as a [Reference](https://aws-controllers-k8s.github.io/community/docs/contributor-docs/code-generator-config/#references-making-a-field-refer-to-another-resource) for the fields `spec.vpcRef` (instead of `spec.vpcId`) and `spec.peerVpcRef` (instead of `spec.peerVpcId`).
- Adds the `spec.accepterPeeringConnectionOptions` and `spec.requesterPeeringConnectionOptions` fields, which come from the `ModifyVpcPeeringConnectionOptions` API call, to allow the use of the `allowDnsResolutionFromRemoteVpc`, `AllowEgressFromLocalVpcToRemoteClassicLink` and `AllowEgressFromLocalClassicLinkToRemoteVpc` sub-fields.
- Adds the `spec.acceptRequest` (boolean) which will allow the VPC Peering Connection Request to automatically switch from status `pending-acceptance` to `active` without manual intervention. (Inspired by Crossplane's implementation ([Ref](https://doc.crds.dev/github.com/crossplane/provider-aws/ec2.aws.crossplane.io/VPCPeeringConnection/v1alpha1@v0.45.1)))
- Adds the field `spec.routes.vpcPeeringConnectionRef` to the Route Table CR

> Note: In [a separate PR](https://github.com/aws-controllers-k8s/ec2-controller/pull/159), we add the fields `spec.acceptVPCPeeringRequestsFromVPCID` and `spec.acceptVPCPeeringRequestsFromVPCRefs` to the VPC CRD, to allow you to add a list of VPC's from which you'd like to automatically accept incoming VPC Peering requests from.

## References
- [Slack Thread](https://kubernetes.slack.com/archives/C0402D8JJS1/p1700682197552809)
- Github Issue: https://github.com/aws-controllers-k8s/community/issues/1943


generator.yaml
```diff
@@ -184,6 +184,10 @@ operations:
     operation_type:
       - Delete
     resource_name: Instance
+  ModifyVpcPeeringConnectionOptions:
+    operation_type:
+      - Update
+    resource_name: VpcPeeringConnection
```
marcdavoli, line 190: Fixes the code-generator mapping of the `sdkUpdate` function to use `ModifyVpcPeeringConnectionOptions` ([Ref](https://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_ModifyVpcPeeringConnectionOptions.html)) instead of `UpdateVpcPeeringConnection` (which doesn't exist)

generator.yaml
```diff
@@ -779,5 +783,31 @@ resources:
...
+          resource: VPC
+          path: Status.VPCID
+      AccepterPeeringConnectionOptions:
+        from:
+          operation: ModifyVpcPeeringConnectionOptions
+          path: AccepterPeeringConnectionOptions
+      RequesterPeeringConnectionOptions:
+        from:
+          operation: ModifyVpcPeeringConnectionOptions
+          path: RequesterPeeringConnectionOptions
```
marcdavoli, line 811: These two fields are used by the `ModifyVpcPeeringConnectionOptions` API call, not the `CreateVpcPeeringConnection` call, but are represented as being natively part of the ACK resource.

apis/v1alpha1/generator.yaml
```diff
@@ -779,5 +783,34 @@ resources:
...
+        from:
+          operation: ModifyVpcPeeringConnectionOptions
+          path: AccepterPeeringConnectionOptions
+      RequesterPeeringConnectionOptions:
+        from:
+          operation: ModifyVpcPeeringConnectionOptions
+          path: RequesterPeeringConnectionOptions
+    hooks:
+      sdk_update_post_build_request:
+        template_path: hooks/vpc_peering_connection/sdk_update_post_build_request.go.tpl
```
marcdavoli, line 814: Wires up 2 fields that exist in the resource's Spec to include them into the Update operation payload
- `AccepterPeeringConnectionOptions.AllowDNSResolutionFromRemoteVPC`
- `RequesterPeeringConnectionOptions.AllowDNSResolutionFromRemoteVPC`

apis/v1alpha1/generator.yaml
```diff
@@ -779,5 +784,33 @@ resources:
...
+  VpcPeeringConnection:
+    fields:      
+      VpcId:
+        references:
+          resource: VPC
+          path: Status.VPCID
+      peerVpcId:
+        references:
+          resource: VPC
+          path: Status.VPCID
```
marcdavoli, line 797: Allows the `vpcId` and `peerVpcId` fields to be replaced with `vpcRef` and `peerVpcRef` fields.

apis/v1alpha1/generator.yaml
```diff
@@ -779,5 +784,33 @@ resources:
...
+      AccepterPeeringConnectionOptions:
+        from:
+          operation: ModifyVpcPeeringConnectionOptions
+          path: AccepterPeeringConnectionOptions
+      RequesterPeeringConnectionOptions:
+        from:
+          operation: ModifyVpcPeeringConnectionOptions
+          path: RequesterPeeringConnectionOptions
```
marcdavoli, line 805: Adds 2 fields to the CRD that are useful to the VPC Peering Connection's `Update` operation.

generator.yaml
```diff
@@ -811,6 +811,4 @@ resources:
           is_ignored: True
     hooks:
       sdk_update_post_build_request:
-        template_path: hooks/vpc_peering_connection/sdk_update_post_build_request.go.tpl
-    tags:
-      ignore: true
```
marcdavoli: Removed this to make the resource's block inline with what all other resources are doing with their `Tags` block:
```yaml
      Tags:
        from:
          operation: CreateTags
          path: Tags
        compare:
          is_ignored: True
```

pkg/resource/vpc_peering_connection/hooks.go
marcdavoli, line 1: Mostly copied from the VPC resource's equivalent file https://github.com/aws-controllers-k8s/ec2-controller/blob/main/pkg/resource/vpc/hooks.go

pkg/resource/vpc_peering_connection/sdk.go
```diff
@@ -0,0 +1,756 @@
...
+	rlog := ackrtlog.FromContext(ctx)
+	exit := rlog.Trace("rm.sdkUpdate")
+	defer func() {
+		exit(err)
+	}()
+	// Custom update function to for Tags
+	desired, err = rm.customUpdateVPCPeeringConnection(ctx, desired, latest, delta)
+	if err != nil {
+		return nil, err
+	}
```
marcdavoli, line 491: This is code that comes from a hook, would like your thoughts on this @a-hilaly .
Thanks

a-hilaly:
```go
if delta.DifferentAt("Tags) { syncTags(...) }

if !delta.DifferentExcept("Tags") {
     return desired, nil
}
```

templates/hooks/vpc_peering_connection/sdk_update_post_build_request.go.tpl
```diff
@@ -0,0 +1,21 @@
...
+	}
+	if desired.ko.Spec.RequesterPeeringConnectionOptions.AllowDNSResolutionFromRemoteVPC == &trueCondition {
+		input.RequesterPeeringConnectionOptions.AllowDnsResolutionFromRemoteVpc = &trueCondition
+	}
+	if desired.ko.Spec.RequesterPeeringConnectionOptions.AllowEgressFromLocalClassicLinkToRemoteVPC == &trueCondition {
+		input.RequesterPeeringConnectionOptions.AllowEgressFromLocalClassicLinkToRemoteVpc = &trueCondition
+	}
+	if desired.ko.Spec.RequesterPeeringConnectionOptions.AllowEgressFromLocalVPCToRemoteClassicLink == &trueCondition {
+		input.RequesterPeeringConnectionOptions.AllowEgressFromLocalVpcToRemoteClassicLink = &trueCondition
+	}
```
a-hilaly, line 21: Let's try to use this https://github.com/aws-controllers-k8s/code-generator/pull/406

pkg/resource/vpc_peering_connection/sdk.go
marcdavoli: Addressed in commit https://github.com/aws-controllers-k8s/ec2-controller/pull/158/commits/2fcdcaa305d640ff8ff20ffa19fddfbdd0a9da44

templates/hooks/vpc_peering_connection/sdk_update_post_build_request.go.tpl
marcdavoli: Addressed in commits:
https://github.com/aws-controllers-k8s/ec2-controller/pull/158/commits/5c1e2eca852f6c0ba2002932cb281fa28c9c2dcf
https://github.com/aws-controllers-k8s/ec2-controller/pull/158/commits/fc5e9db3baf1c602d936733ca60e0240964e308c

marcdavoli: I believe I hit a small bug in the code generator, caused by the update_operation.omit_unchanged_fields field:
https://github.com/aws-controllers-k8s/ec2-controller/commit/5c1e2eca852f6c0ba2002932cb281fa28c9c2dcf#r133988514

pkg/resource/vpc_peering_connection/sdk.go
```diff
@@ -0,0 +1,762 @@
...
+			acceptInput := &svcsdk.AcceptVpcPeeringConnectionInput{
+				VpcPeeringConnectionId: latest.ko.Status.VPCPeeringConnectionID,
+			}
+			acceptResp, err = rm.sdkapi.AcceptVpcPeeringConnectionWithContext(ctx, acceptInput)
+			if err != nil {
+				return nil, err
+			}
+			rlog.Debug("VPC Peering Connection accepted", "VpcPeeringConnectionId", *acceptResp.VpcPeeringConnection.VpcPeeringConnectionId)
+		}
+	}
```
a-hilaly, line 513: Few comments:
- `*desired.ko.Spec.AcceptRequest` will panic if `AcceptRequest` is `nil`
- We need to also check for `delta.DifferentAt("Spec.AcceptRequest")`
- However, for the above, we need a way to get the right "observed" state, so that the controller can react properly. This means we need to add a hook to the `sdkFind` that will be responsible of populating the `spec.AcceptRequest` with latest observed state.

a-hilaly: For point number two, this means that we have to move this block right after L492 and change L495 into `!delta.DifferentExcept("Spec.Tags", "Spec.AcceptRequest")`

marcdavoli:
weird behavior from the tests, I haven't changed any code in those folders in this PR but there are Transit Gateway and Security Group errors in the Kind tests, I will retest to find out if this is just a weird transient error.
/test ec2-kind-e2e

test/e2e/tests/test_vpc_peering_connection.py
```diff
@@ -0,0 +1,341 @@
...
+        resource_name, namespace="default",
+    )
+    k8s.create_custom_resource(ref, resource_data)
+    time.sleep(CREATE_WAIT_AFTER_SECONDS)
+
+    cr = k8s.wait_resource_consumed_by_controller(ref)
+    assert cr is not None
+    assert k8s.get_resource_exists(ref)
+    # Can't uncomment this line until ACK VPCs support auto-accepting VPC Peering Requests
+    # assert cr["status"]["status"]["code"] == "active"
```
a-hilaly, line 190: I guess we can uncomment this now?

test/e2e/tests/test_vpc_peering_connection.py
a-hilaly, line 1: Should we test the case, where we flip the `AcceptRequest` and watch the magic happen?

a-hilaly (commented):
Great work on this @marcdavoli ! I left some comment in-line

marcdavoli:
@a-hilaly , Still getting a ton of weird errors in the tests. Is it possible there is a problem in the backend, such as Quotas being hit in the Testing AWS account?
Thanks

templates/hooks/vpc_peering_connection/sdk_create_post_set_output.go.tpl
```diff
@@ -0,0 +1,14 @@
...
+        _ = acceptResp
+        acceptInput := &svcsdk.AcceptVpcPeeringConnectionInput{
+            VpcPeeringConnectionId: ko.Status.VPCPeeringConnectionID,
+        }
+        acceptResp, err = rm.sdkapi.AcceptVpcPeeringConnectionWithContext(ctx, acceptInput)
+        if err != nil {
+            return nil, err
+        }
+        rlog.Debug("VPC Peering Connection accepted", "VpcPeeringConnectionId", *acceptResp.VpcPeeringConnection.VpcPeeringConnectionId)
+	}
```
a-hilaly, line 14: There is a hack, to trigger and second reconciliation right after a create by using (so you won't have to write any logic here and just let the sdkUpdate deal with it)
```go
	// This causes a requeue and the rest of the fields will be synced on the next
	// reconciliation loop
	ackcondition.SetSynced(&resource{ko}, corev1.ConditionFalse, nil, nil)
```

templates/hooks/vpc_peering_connection/sdk_update_pre_build_request.go.tpl
```diff
@@ -0,0 +1,27 @@
...
+			acceptInput := &svcsdk.AcceptVpcPeeringConnectionInput{
+				VpcPeeringConnectionId: latest.ko.Status.VPCPeeringConnectionID,
+			}
+			acceptResp, err = rm.sdkapi.AcceptVpcPeeringConnectionWithContext(ctx, acceptInput)
+			if err != nil {
+				return nil, err
+			}
+			rlog.Debug("VPC Peering Connection accepted", "VpcPeeringConnectionId", *acceptResp.VpcPeeringConnection.VpcPeeringConnectionId)
+		}
+	}
```
a-hilaly, line 22: This needs to be wrapped with
```
    if delta.DifferentAt("Spec.AcceptRequest") && desired.ko.Spec.AcceptRequest != nil && *desired.ko.Spec.AcceptRequest  {
         ....
	}
```

a-hilaly: Now what we could do is, in case if the user sets the `AcceptRequest` to `true` then `false` - we can show a TerminalCondition in the resource saying that's an "illegal move"... Setting TerminalConditions will prevent the resource from being reconciled unless if they are modified. So at the end this block could be come
```go
    if delta.DifferentAt("Spec.AcceptRequest") {
         if desired.ko.Spec.AcceptRequest == nil || !*desired.ko.Spec.AcceptRequest  {
         		msg := fmt.Sprintf("You cannot set AcceptRequest to false after setting it to true")
		        return nil, ackerr.NewTerminalError(fmt.Errorf(msg))
         }
         // We arrive here only if `*desired.ko.Spec.AcceptRequest` is not nil and true (which is safe to access at this point
         acceptInput := &svcsdk.AcceptVpcPeeringConnectionInput{
             VpcPeeringConnectionId: latest.ko.Status.VPCPeeringConnectionID,
         }
         acceptResp, err = rm.sdkapi.AcceptVpcPeeringConnectionWithContext(ctx, acceptInput)
         if err != nil {
             return nil, err
         }
         rlog.Debug("VPC Peering Connection accepted", "VpcPeeringConnectionId", *acceptResp.VpcPeeringConnection.VpcPeeringConnectionId)
	}


```

a-hilaly: This means that `delta.DifferentAt("Spec.AcceptRequest")` should work, and to do that we might to add at the end of `sdkFind`
```
if ko.Status.Status != nil &&
    ko.Status.Status.Code != nil &&
    *ko.Status.Status.Code == "pending-acceptance" {
    ko.Spec.AcceptRequest = aws.Bool(true)
else {
    ko.Spec.AcceptRequest = aws.Bool(false)
}

```

a-hilaly (commented):
Getting closer 👍 , just tweaking a bit the logic to be similar to what's done in other controllers too.

templates/hooks/vpc_peering_connection/sdk_create_post_set_output.go.tpl
marcdavoli: So I can entirely remove this hook?

templates/hooks/vpc_peering_connection/sdk_update_pre_build_request.go.tpl
marcdavoli: Good idea for the Terminal Error, I added that in commit https://github.com/aws-controllers-k8s/ec2-controller/pull/158/commits/cd7b39b458a302626540e6d52656a8805ea1695d

However, the `spec.acceptRequest` field is a boolean and its default zero-value is `false`, so:
1. We can't do nil comparisons like `if field == nil ...`
2. That would make your code above look like:
  ```
if *ko.Status.Status.Code == "pending-acceptance" {
    ko.Spec.AcceptRequest = aws.Bool(true)
}
  ```
and I don't quite understand why we need to do this?...

a-hilaly: We need this because, we want the controller to trigger an update and more specifically see a difference in `Spec.AcceptRequest`. Overall the logic will be like:
- 1. User submit desired state (with acceptRequest true)
- 2. Controller copy the desired and reads the observed state from AWS
- 3. Controller merges the observed state into the desired state. 
- 4. Controller calls `delta.DifferentAt("Spec")`

In step two we need to set the correct observed state of `AcceptRequest` after an `sdkFind` (step 2)

a-hilaly (approved):
Excellent work on this! Mercy buckets for your contributions!! 👍 🙇 
/approve
/lgtm

a-hilaly (approved):
Allez, cette fois, c'est la bonne!
Mercy buckets!!
/lgtm
/approve

PR #160, opened 2023-12-05
Add support for adoption of existing Flow Logs and Security Groups

pcolazurdo:
Issue #, if available: https://github.com/aws-controllers-k8s/community/issues/1946

Description of changes:

This proposed fix add support for additionalKeys needed to support adoption of Security Groups and FlowLogs as requested on the issue linked above.

The following adoption instructions will now work:

```yaml
apiVersion: services.k8s.aws/v1alpha1
kind: AdoptedResource
metadata:
  name: my-flow-log
spec:
  aws:
    nameOrID: fl-xxxxxxx
    additionalKeys:
      resourceID: vpc-xxxxxx
      resourceType: VPC
  kubernetes:
    group: ec2.services.k8s.aws
    kind: FlowLog
---
apiVersion: services.k8s.aws/v1alpha1
kind: AdoptedResource
metadata:
  name: my-sg
spec:
  aws:
    nameOrID: sg-xxxxxx
    additionalKeys:
      name: "My GroupName"
  kubernetes:
    group: ec2.services.k8s.aws
    kind: SecurityGroup
```

### DIFF
--- a/templates/hooks/flow_log/post_set_resource_identifiers.go.tpl
+++ b/templates/hooks/flow_log/post_set_resource_identifiers.go.tpl
@@ -1,0 +1,8 @@
+	f6, f6ok := identifier.AdditionalKeys["resourceID"]
+	if f6ok {
+		r.ko.Spec.ResourceID = &f6
+	}
+        f7, f7ok := identifier.AdditionalKeys["resourceType"]
+        if f7ok {
+                r.ko.Spec.ResourceType = &f7
+        }
--- a/templates/hooks/flow_log/post_set_resource_identifiers.go.tpl
+++ b/templates/hooks/flow_log/post_set_resource_identifiers.go.tpl
@@ -1,8 +1,8 @@
-	f6, f6ok := identifier.AdditionalKeys["resourceID"]
-	if f6ok {
-		r.ko.Spec.ResourceID = &f6
+	identifierResourceID, identifierResourceIDOk := identifier.AdditionalKeys["resourceID"]
+	if identifierResourceIDOk {
+		r.ko.Spec.ResourceID = &identifierResourceID
 	}
-        f7, f7ok := identifier.AdditionalKeys["resourceType"]
-        if f7ok {
-                r.ko.Spec.ResourceType = &f7
+        identifierResourceType, identifierResourceTypeOk := identifier.AdditionalKeys["resourceType"]
+        if identifierResourceTypeOk {
+                r.ko.Spec.ResourceType = &identifierResourceType
         }
--- a/templates/hooks/security_group/post_set_resource_identifiers.go.tpl
+++ b/templates/hooks/security_group/post_set_resource_identifiers.go.tpl
@@ -1,4 +1,4 @@
-	f6, f6ok := identifier.AdditionalKeys["name"]
-	if f6ok {
-		r.ko.Spec.Name = &f6
+	identifierName, identifierNameOk := identifier.AdditionalKeys["name"]
+	if identifierNameOk {
+		r.ko.Spec.Name = &identifierName
 	}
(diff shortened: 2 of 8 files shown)


templates/hooks/flow_log/post_set_resource_identifiers.go.tpl
```diff
@@ -0,0 +1,8 @@
+	f6, f6ok := identifier.AdditionalKeys["resourceID"]
+	if f6ok {
+		r.ko.Spec.ResourceID = &f6
+	}
+        f7, f7ok := identifier.AdditionalKeys["resourceType"]
+        if f7ok {
+                r.ko.Spec.ResourceType = &f7
+        }
```
a-hilaly, line 8: We can use non-numbered variables here:
```

	resourceID, ok := identifier.AdditionalKeys["resourceID"]
	if ok {
		r.ko.Spec.ResourceID = &resourceID
	}
    resourceType, ok := identifier.AdditionalKeys["resourceType"]
    if ok {
        r.ko.Spec.ResourceType = &resourceType
    }
```

templates/hooks/security_group/post_set_resource_identifiers.go.tpl
```diff
@@ -0,0 +1,4 @@
+	f6, f6ok := identifier.AdditionalKeys["name"]
+	if f6ok {
+		r.ko.Spec.Name = &f6
+	}
```
a-hilaly, line 4: Same comment

a-hilaly (commented):
Neat, thank you @pcolazurdo ! i left few comments in-line
Also can we add some e2e tests to validate that those adoptions are working and won't break in the future?

templates/hooks/flow_log/post_set_resource_identifiers.go.tpl
pcolazurdo: fixed

templates/hooks/security_group/post_set_resource_identifiers.go.tpl
pcolazurdo: fixed

pcolazurdo:
Tried to find an example of an adopted check to use as a baseline, but couldn't find one. For these two resources adoption, we should create a new Security Group manually, and create a new VPC with flow log enabled. How should I proceed with this? I don't want to stand up new resources that will be hard to tear down after, and I don't fully understand the e2e workflow.

a-hilaly:
> Tried to find an example of an adopted check to use as a baseline, but couldn't find one. For these two resources adoption, we should create a new Security Group manually, and create a new VPC with flow log enabled. How should I proceed with this? I don't want to stand up new resources that will be hard to tear down after, and I don't fully understand the e2e workflow.

Instead of creating everything manually, you can use ACK it self to provision those resources... once provisioned you can annotate those resources with a delete-retain-policy and try to adopt right after. https://aws-controllers-k8s.github.io/community/docs/user-docs/deletion-policy/

a-hilaly:
@pcolazurdo Did you test those changes locally? happy to merge if you say it worked for locally. We need a global plan to add resource adoption tests..

pcolazurdo:
I've done some more testing and there is an issue with the Security Group adoption that I haven't been able to solve as this seems to be an issue with this particular API and how the runtime manager required fields. 
The name property in the Security Group Resource is a required property for creation, but because it can be mutated later, it shouldn't be considered as an Identity property. The problem seem to be that the existing specs doesn't let you define a property as required for creation but not for reading (ReadOne). Defining the Name property as an additionalKey for adoption produce the side effect that if the Name specified in the adoption CRD is different that the existing on AWS, the property will be mutated, which is undesired side effect that can break existing environments. We need to find a way to specify properties as required for creation but optional for read.
I'll write a but report on the runtime repo with this

I'll also create a different PR for the VPC Flow Logs piece. In this case, the AWS API doesn't allow mutation so there is no side effect if the user makes a mistake when defining the values of the additionalKeys.

PR #164, opened 2023-12-12
Fix `AdoptedResource` functionality for `SecurityGroups`

nnbu:
Issue # : https://github.com/aws-controllers-k8s/community/issues/1946

Description of changes:
RCA:
Following CR currently fails to adopt the existing SecurityGroups resource.

```
apiVersion: services.k8s.aws/v1alpha1
kind: AdoptedResource
metadata:
  name: sg1
spec:
  aws:
    nameOrID: sg-07fd54f8e16eec74a
  kubernetes:
    group: ec2.services.k8s.aws
    kind: SecurityGroup
    metadata:
      name: sg-adopted
      namespace: default
```

The error
```
"error": "SecurityGroup.ec2.services.k8s.aws \"my-resource\" is invalid: spec.name: Required value"
```
comes from https://github.com/aws-controllers-k8s/runtime/blob/main/pkg/runtime/adoption_reconciler.go#L157

SetIdentifiers function needs to have code to fill up required values of the CR so as to solve this issue.

Solution:
A new fields is defined as name and it is initialized in SetIdentifiers function. This field needs to be part of additionalKeys field of AdoptedResource CR for SecurityGroups.

The new CR looks like,

```
apiVersion: services.k8s.aws/v1alpha1
kind: AdoptedResource
metadata:
  name: sg1
spec:
  aws:
    nameOrID: sg-xxxxxx
    additionalKeys:
      name: <security-group-name>
  kubernetes:
    group: ec2.services.k8s.aws
    kind: SecurityGroup
    metadata:
      name: sg-adopted
      namespace: default
```

### DIFF
--- a/apis/v1alpha1/ack-generate-metadata.yaml
+++ b/apis/v1alpha1/ack-generate-metadata.yaml
@@ -1,13 +1,13 @@
 ack_generate_info:
-  build_date: "2023-12-06T21:43:43Z"
-  build_hash: 892f29d00a4c4ad21a2fa32919921de18190979d
+  build_date: "2023-12-13T07:19:26Z"
+  build_hash: 3653329ceeb20015851b8776a6061a3fb0ec2935
   go_version: go1.21.1
-  version: v0.27.1
-api_directory_checksum: 6e2d850d97f2f72db31c9bef522eca4ab95b3fcd
+  version: v0.27.1-6-g3653329
+api_directory_checksum: d452bf19bfd1496aacdc215bf7cc9ea86c55c122
 api_version: v1alpha1
 aws_sdk_go_version: v1.44.93
 generator_config_info:
-  file_checksum: d10a62517f87bacf988184f8c454b90b42dee732
+  file_checksum: ff084526a7037be1c7d08e2c7e742ac0679abc5f
   original_file_name: generator.yaml
 last_modification:
   reason: API generation
--- a/apis/v1alpha1/generator.yaml
+++ b/apis/v1alpha1/generator.yaml
@@ -549,6 +549,8 @@ resources:
         template_path: hooks/security_group/sdk_create_post_set_output.go.tpl
       sdk_read_many_post_set_output:
         template_path: hooks/security_group/sdk_read_many_post_set_output.go.tpl
+      post_set_resource_identifiers:
+        template_path: hooks/security_group/post_set_resource_identifiers.go.tpl
     update_operation:
       custom_method_name: customUpdateSecurityGroup
   NetworkAcl:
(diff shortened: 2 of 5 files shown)


nnbu:
Hi @a-hilaly 
This change is similar to FlowLog change which was committed yesterday. Could you please review?

pcolazurdo:
Hi @nnbu - have you tested this? I've tried something similar [before](https://github.com/aws-controllers-k8s/ec2-controller/pull/160), but in this case, if the user set the wrong Name in the adoption, the existing value in AWS is overwritten which is an undesired side effect of this. Just checking if this is not the case. I have a PR in progress that follows a different method and avoid that problem

nnbu:
> Hi @nnbu - have you tested this? I've tried something similar [before](https://github.com/aws-controllers-k8s/ec2-controller/pull/160), but in this case, if the user set the wrong Name in the adoption, the existing value in AWS is overwritten which is an undesired side effect of this. Just checking if this is not the case. I have a PR in progress that follows a different method and avoid that problem

Hi [pcolazurdo](https://github.com/pcolazurdo)
I have tested this and I dont see the behavior you mentioned.

Here is what I have tried.

1. Created SG from aws console with `security group name` as `test-1`
2. Then created following CR
```
apiVersion: services.k8s.aws/v1alpha1
kind: AdoptedResource
metadata:
  name: sg1
spec:
  aws:
    nameOrID: <sg created in the previous step>
    additionalKeys:
      name: nonexisting-name
  kubernetes:
    group: ec2.services.k8s.aws
    kind: SecurityGroup
    metadata:
      name: sg-adopted1
      namespace: default
```

3. Verified in the console that `security group name` remains unchanged. 
4. Of course, when this wrong name is specified in the CR, there is a mismatch between `security group name` in the CR and that seen from aws console. But that does not affect adoptability.

nnbu:
In fact, there is an unrelated issue here.
`name` change in securityGroup CR does not change the name of aws SG resource.

Here are the steps
1. Apply following CR
```
kind: SecurityGroup
metadata:
  name: sg-not-adopted1
  namespace: default
spec:
  description: not-adopted
  egressRules:
  - ipProtocol: "-1"
    ipRanges:
    - cidrIP: 0.0.0.0/0
  name: not-adopted-sg
  vpcID: vpc-xxxxxx
```

2. Verify in the aws console that `security group name` is set to `not-adopted-sg`.
3. change the name in the CR so that it looks like following
```
kind: SecurityGroup
metadata:
  name: sg-not-adopted1
  namespace: default
spec:
  description: not-adopted
  egressRules:
  - ipProtocol: "-1"
    ipRanges:
    - cidrIP: 0.0.0.0/0
  name: not-adopted-sg-CHANGED
  vpcID: vpc-xxxxxx
```

4. The `security group name` is not changed in aws console.

This happens because ec2 controller does not detect the name change when it calculates the `diff` between `desired` CR and `existing` resource. 

This bug is not related to `adoptResrouce` functionality though.

a-hilaly:
implemented in https://github.com/aws-controllers-k8s/ec2-controller/pull/165

PR #169, opened 2023-12-14
networkAcl update fails if ruleAction field has camelcase Allow/Deny

nnbu:
Issue [#1966](https://github.com/aws-controllers-k8s/community/issues/1966)

Description of changes:
RCA and steps to reproduce are mentioned in the issue

### DIFF
--- a/pkg/resource/network_acl/hooks.go
+++ b/pkg/resource/network_acl/hooks.go
@@ -17,6 +17,7 @@ import (
 	"context"
 	"errors"
 	"strconv"
+	"strings"
 
 	ackcompare "github.com/aws-controllers-k8s/runtime/pkg/compare"
 	ackrtlog "github.com/aws-controllers-k8s/runtime/pkg/runtime/log"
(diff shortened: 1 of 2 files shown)


pkg/resource/network_acl/sdk.go
```diff
@@ -615,7 +615,7 @@ func compareNetworkACLEntry(
 	if ackcompare.HasNilDifference(a.RuleAction, b.RuleAction) {
 		delta.Add("NetworkACLEntry.RuleAction", a.RuleAction, b.RuleAction)
 	} else if a.RuleAction != nil && b.RuleAction != nil {
-		if *a.RuleAction != *b.RuleAction {
+		if !strings.EqualFold(*a.RuleAction, *b.RuleAction) {
```
a-hilaly, line 618: @nnbu The templates modifications are missing from this PR. Can you please include them?
/ok-to-test

nnbu: @a-hilaly 
`compareNetworkACLEntry` is not custom function. is it possible to make template modification for this type? 

I agree that we cant just make this change, as it would get overwritten.

a-hilaly: @nnbu I would say that it's possible.. if not we can make it a hook function

pkg/resource/network_acl/hooks.go
```diff
@@ -403,6 +406,13 @@ func containsEntry(
 	}
 	return false
 }
+
+func toLowercase(str *string) {
+	if str != nil {
+		*str = strings.ToLower(*str)
+	}
+}
```
a-hilaly, line 414: This will cause the desired RuleAction to change, and later patched back by the controller, which can trigger infinite reconciliations if you're using GitOps tools.. Can we consider comparing a copies of those rules?

nnbu: The problem is that, `compareNetworkACLEntry` is a generated function which I cant modify. So, I am trying to modify the arguments passed to it. 
I can copy entire `e` and `entry` and modify those copies as above and pass to `compareNetworkACLEntry`

nnbu: I updated the logic now.

a-hilaly (approved):
👍 
/lgtm

PR #191, opened 2024-05-21
Fix: EC2 controller for vpcendpoint does not update status

nnbu:
Issue : https://github.com/aws-controllers-k8s/community/issues/2075

Description of changes:
The issue is that vpce CR status is not updated. This happens because when vpce is created, it takes around a minute for it to be created in aws. Due to this, it's status is kept `pending`. But there is no sync enforced on the reconciler. Due to this, reconciliation happens after 10h default time. At this time, status would be set correctly.

This fix enforces reconciliation to make sure CR is updated correctly when vpce creation in aws is successful. 

### DIFF
--- a/apis/v1alpha1/ack-generate-metadata.yaml
+++ b/apis/v1alpha1/ack-generate-metadata.yaml
@@ -1,13 +1,13 @@
 ack_generate_info:
-  build_date: "2024-05-02T20:55:46Z"
+  build_date: "2024-05-22T16:42:26Z"
   build_hash: 14cef51778d471698018b6c38b604181a6948248
-  go_version: go1.22.0
+  go_version: go1.21.1
   version: v0.34.0
-api_directory_checksum: 7fd395ceb7d5d8e35906991c7348d3498f384741
+api_directory_checksum: 1b53401670898ce50e6d6cc8bfba6b63ea7d5683
 api_version: v1alpha1
 aws_sdk_go_version: v1.44.93
 generator_config_info:
-  file_checksum: 21aa2c0035772a834102dcdac81a186fe148212c
+  file_checksum: 75820b9d685b38976cd9723a9435f119ab913245
   original_file_name: generator.yaml
 last_modification:
   reason: API generation
--- a/templates/hooks/vpc_endpoint/sdk_read_many_post_set_output.go.tpl
+++ b/templates/hooks/vpc_endpoint/sdk_read_many_post_set_output.go.tpl
@@ -1,0 +1,8 @@
+
+	if !vpcEndpointAvailable(&resource{ko}) {
+		// Setting resource synced condition to false will trigger a requeue of
+		// the resource. No need to return a requeue error here.
+		ackcondition.SetSynced(&resource{ko}, corev1.ConditionFalse, nil, nil)
+	} else {
+		ackcondition.SetSynced(&resource{ko}, corev1.ConditionTrue, nil, nil)
+	}
--- a/apis/v1alpha1/generator.yaml
+++ b/apis/v1alpha1/generator.yaml
@@ -823,8 +823,12 @@ resources:
         code: compareTags(delta, a, b)
       sdk_create_post_build_request:
         template_path: hooks/vpc_endpoint/sdk_create_post_build_request.go.tpl
+      sdk_create_post_set_output:
+        template_path: hooks/vpc_endpoint/sdk_create_post_set_output.go.tpl
       sdk_delete_post_build_request:
         template_path: hooks/vpc_endpoint/sdk_delete_post_build_request.go.tpl
+      sdk_read_many_post_set_output:
+        template_path: hooks/vpc_endpoint/sdk_read_many_post_set_output.go.tpl
       sdk_file_end:
         template_path: hooks/vpc_endpoint/sdk_file_end.go.tpl
     update_operation:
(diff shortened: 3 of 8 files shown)


a-hilaly (approved):
This is great, thank you @nnbu !
Shall we make a release for this?

apis/v1alpha1/ack-generate-metadata.yaml
```diff
@@ -1,13 +1,13 @@
 ack_generate_info:
-  build_date: "2024-05-02T20:55:46Z"
-  build_hash: 14cef51778d471698018b6c38b604181a6948248
-  go_version: go1.22.0
-  version: v0.34.0
+  build_date: "2024-05-21T22:16:46Z"
+  build_hash: d660ee36fe947607ebea039acd47c35477b4a836
+  go_version: go1.21.1
+  version: v0.28.0-58-gd660ee3
```
a-hilaly, line 5: Can please you use the latest version of code-gen? 
/hold

a-hilaly:
@nnbu can you also update the e2e tests? i think we might have to wait more than 5x30 seconds
/retest

apis/v1alpha1/ack-generate-metadata.yaml
nnbu: done

nnbu:
> @nnbu can you also update the e2e tests? i think we might have to wait more than 5x30 seconds /retest

@a-hilaly this is done now

a-hilaly (approved):
Thank you @nnbu !
/lgtm
/test all

wjf3121:
@nnbu @a-hilaly 

I found a potential bug with this implementation: https://github.com/aws-controllers-k8s/community/issues/1933#issuecomment-2149307339

It seems that we stop syncing the state once the VPCEndpoint becomes `available`. However an `available` VPCEndpoint can still later be rejected and the controller will not update the state accordingly.

nnbu:
The fix addresses the issue for the case when the vpc endpoint goes from
`pending` to `available`.

Your latest comment is about the case when the state changes from
`available` to `rejected`. As you mentioned, the trigger for this
transition is manual trigger which is not managed/controlled by ack
controller.

It is not possible to handle such cases without reconciling every single
resource at a sufficiently high frequency. E.g. consider that if we
delete/change any resource out of the scope of ack controllers, then it
won't be reflected in ack immediately.

Ack has a default reconciling frequency of 10 hours. So, only at that
frequency, ack detects the change of state of the resource.

templates/hooks/vpc_endpoint/sdk_read_many_post_set_output.go.tpl
```diff
@@ -0,0 +1,8 @@
+
+	if !vpcEndpointAvailable(&resource{ko}) {
+		// Setting resource synced condition to false will trigger a requeue of
+		// the resource. No need to return a requeue error here.
+		ackcondition.SetSynced(&resource{ko}, corev1.ConditionFalse, nil, nil)
+	} else {
+		ackcondition.SetSynced(&resource{ko}, corev1.ConditionTrue, nil, nil)
+	}
```
wjf3121, line 8: @nnbu Thanks for the explanation [in this comment](https://github.com/aws-controllers-k8s/ec2-controller/pull/191#issuecomment-2149384224), it makes sense to me mostly! I still have one question over this implmenetation.


Here we consider a `VPCEndpoint` resource not synced as long as it's not in `available` state. However a VPC endpoint can be in `pendingAcceptance` state after creation and stay in that state forever/later be in `rejected` state.

It seems to me that in controller's perspective it should be considered `synced` in these cases. In earlier versions of the controller that indeed is that case. But with this change, the VPCEndpoint will remain unsynced  after its creation unless it becomes `available` later, which may never happen in some cases. 

Does it make sense to only force the resource to be unsynced only when the resource is in `pending` instead of when it's in a non `available` state? This seems more inline with the [endpoint states definition](https://docs.aws.amazon.com/vpc/latest/privatelink/concepts.html#concepts-endpoint-states):

> Pending - The service provider accepted the connection request. This is the initial state if requests are automatically accepted. The VPC endpoint returns to this state if the service consumer modifies the VPC endpoint.

wjf3121: We have such automated pipeline that for a VPCEndpoint that requires manual acceptance, we first wait til the resource to be synced and then accept the connection through API. The change will break the behavior.

wjf3121: @nnbu @a-hilaly , if it makes sense to you, here's the fix I proposed on top of this change: https://github.com/aws-controllers-k8s/ec2-controller/pull/198

nnbu:
>> Does it make sense to only force the resource to be unsynced only when
the resource is in pending instead of when it's in a non available state?

This makes sense totally. What are some of the transitionary states for vpc
endpoint, similar to pending? We should basically treat all such
transitionary states as `notSynced`.


@amine will be able to check with the service teams and get such states.
Depending on that, we can make the change. Your PR looks good to me if
`pending` is the only transitionary state.

wjf3121:
> Your PR looks good to me if `pending` is the only transitionary state.

`pending` and `deleting` are the only two transitionary states AFAIK and `deleting` should already be handled correctly assuming the resource is fully managed by the controller. It would be awesome if you can confirm this internally:)

PR #203, opened 2024-06-27
fix(securityGroup): groupRef in userIDGroupPairs references id instead of name

kejne:
Handles issues [#2101](https://github.com/aws-controllers-k8s/community/issues/2101)

GroupId is populated instead of the groupName (which would be the same as the name passed in as reference) when using the groupRef solution.


a-hilaly (approved):
Great catch, thank you very much @kejne !

/lgtm

a-hilaly:
@kejne mind shipping a release PR? something like `RELEASE_VERSION=v1.2.12 ./scripts/build-controller-release.sh ec2` should do the job

PR #212, opened 2024-08-19
Fix: Add option to delete default SG rule created upon VPC creation

nnbu:
Issue #, if available:

Description of changes:
When a VPC is created, it also creates 'default' security group. The 'default' security group has widely open egress rules. We need to have an option to delete this autocreated rules from 'default' security group. The 'default' security group itself can not be deleted.

Discussion:
https://kubernetes.slack.com/archives/C0402D8JJS1/p1720560499642019


pkg/resource/vpc/sdk.go
```diff
@@ -207,6 +207,15 @@ func (rm *resourceManager) sdkFind(
 		ko.Spec.EnableDNSSupport = dnsAttrs.EnableSupport
 		ko.Spec.EnableDNSHostnames = dnsAttrs.EnableHostnames
 	}
+	defaultSGRuleExist, err := rm.hasDefaultSecurityGroupRule(ctx, &resource{ko})
```
a-hilaly, line 210: We need to handle this error

pkg/resource/vpc/hooks.go
```diff
@@ -350,3 +363,169 @@ func updateTagSpecificationsInCreateRequest(r *resource,
...
+		},
+	}
+
+	resp, err := rm.sdkapi.DescribeSecurityGroupsWithContext(ctx, input)
+	rm.metrics.RecordAPICall("READ_MANY", "DescribeSecurityGroupRules", err)
+	if err != nil || resp == nil {
+		return nil, err
+	}
+
+	return resp.SecurityGroups[0].GroupId, nil
```
a-hilaly, line 522: I guess it's safe to access [0] here, since default SG can't be deleted?

pkg/resource/vpc/hooks.go
```diff
@@ -350,3 +363,169 @@ func updateTagSpecificationsInCreateRequest(r *resource,
...
+}
+
+func toStrPtr(str string) *string {
+	return &str
+}
+
+func toInt64Ptr(integer int64) *int64 {
+	return &integer
+}
```
a-hilaly, line 531: nit:
```go
func ptr[t T](target T) *T { return &t }
```

pkg/resource/vpc/hooks.go
```diff
@@ -350,3 +363,169 @@ func updateTagSpecificationsInCreateRequest(r *resource,
...
+		return false
+	}
+
+	if *rule.CidrIpv4 == "0.0.0.0/0" &&
+		*rule.FromPort == -1 &&
+		*rule.ToPort == -1 &&
+		*rule.IpProtocol == "-1" &&
+		*rule.IsEgress {
+		return true
+	}
```
a-hilaly, line 443: i wonder if non default security groups can have these attributes? i guess no, just making sure we're getting it right

pkg/resource/vpc/hooks.go
```diff
@@ -293,6 +294,18 @@ func (rm *resourceManager) customUpdateVPC(
...
+		if err = rm.deleteDefaultSecurityGroupRule(ctx, desired); err != nil {
+			// if deleteDefaultSecurityGroupRule fails, assume that the rule
+			// still exists and update the status accordingly.
+			exist := true
+			updated.ko.Status.DefaultSecurityGroupRuleExist = &exist
+			return nil, err
+		}
+		exist := false
+		updated.ko.Status.DefaultSecurityGroupRuleExist = &exist
+	}
```
a-hilaly, line 307: Do we want to wrap this with `if delta.DifferentAt("Spec.DeleteDefaultSecurityGroupRule")` ?

test/e2e/tests/test_vpc.py
```diff
@@ -188,6 +229,74 @@ def test_crud_tags(self, ec2_client, simple_vpc):
...
+        ec2_validator.assert_vpc(resource_id)
+
+        # Make sure default security group rule is not deleted
+        assert contains_default_sg_rule(ec2_client, resource_id)
+
+        # Set deleteDefaultSecurityGroupRule to delete default security group
+        # rule
+        updates = {
+            "spec": {"deleteDefaultSecurityGroupRule": True}
+        }
```
a-hilaly, line 272: This feels a little like an imperative call, instead of a declarative approach.. especially when switching back to False is a no-op. Just thinking out-loud, i'm not against this approach.

a-hilaly (commented):
Good stuff, thank you so much @nnbu for your contribution! :) 
i left few comments in-line

pkg/resource/vpc/hooks.go
nnbu: `isDefaultSGEgressRule` is only checking whether the passed rule is same as auto populated egress rule or not. Non-default security groups can intentionally add this rule, if they need. However, `isDefaultSGEgressRule` function gets called from `hasDefaultEgressRule` which gets security group ID of 'default' security group. So, we are never going to touch this rule for non-default security groups.

test/e2e/tests/test_vpc.py
nnbu: My intentions was to test two scenarios.
1. In test_delete_default_security_group_rule : set `deleteDefaultSecurityGroupRule` to `true` upfront. This tests `create` path
2. In test_update_delete_default_security_group_rule : set `deleteDefaultSecurityGroupRule` to `true` as part of Update. This tests `update` path

pkg/resource/vpc/sdk.go
nnbu: done

pkg/resource/vpc/hooks.go
nnbu: yes, default SG has to be present. But I guess, it does not hurt checking for the access. So, I added it now

pkg/resource/vpc/hooks.go
nnbu: done. Thanks

pkg/resource/vpc/hooks.go
nnbu: done

pkg/resource/vpc/hooks.go
```diff
@@ -350,3 +366,217 @@ func updateTagSpecificationsInCreateRequest(r *resource,
...
+		*rule.IpProtocol == "-1" &&
+		*rule.IsEgress {
+		return true
+	}
+	return false
+}
+
+// deleteDefaultSecurityGroupRule deletes the egress rule that is attached to
+// the 'default' SecurityGroup upon creation.
+func (rm *resourceManager) deleteDefaultSecurityGroupRule(
```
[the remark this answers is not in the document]
nnbu: done

pkg/resource/vpc/hooks.go
```diff
@@ -350,3 +366,217 @@ func updateTagSpecificationsInCreateRequest(r *resource,
...
+		*rule.ToPort == -1 &&
+		*rule.IpProtocol == "-1" &&
+		*rule.IsEgress {
+		return true
+	}
+	return false
+}
+
+// deleteDefaultSecurityGroupRule deletes the egress rule that is attached to
+// the 'default' SecurityGroup upon creation.
```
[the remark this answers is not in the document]
nnbu: done

pkg/resource/vpc/hooks.go
```diff
@@ -350,3 +366,217 @@ func updateTagSpecificationsInCreateRequest(r *resource,
...
+
+	return rm.hasDefaultRules(ctx, sgID)
+}
+
+// hasDefaultRules calls DescribeSecurityGroupRules and returns true if
+// the security group has autopopulated ingress/egress rule.
+func (rm *resourceManager) hasDefaultRules(
+	ctx context.Context,
+	sgID *string,
+) (isDefaultSGEgressRulePresent bool, err error) {
```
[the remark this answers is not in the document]
nnbu: done

pkg/resource/vpc/hooks.go
```diff
@@ -350,3 +366,217 @@ func updateTagSpecificationsInCreateRequest(r *resource,
 		input.TagSpecifications = []*svcsdk.TagSpecification{&desiredTagSpecs}
 	}
 }
+
+// hasDefaultSecurityGroupRule returns true if the vpc's 'default' security
+// group has autopopulated egress rule.
```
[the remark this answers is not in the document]
nnbu: done

apis/v1alpha1/generator.yaml
```diff
@@ -725,6 +725,12 @@ resources:
         print:
           path: Status.vpcID
           name: ID
+      DisallowSecurityGroupDefaultRules:
+        type: bool
+        is_required: false
+      DefaultSecurityGroupRuleExist:
```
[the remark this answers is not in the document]
nnbu: I agree. I have now renamed everything to on the lines of `SecurityGroupDefaultRules`

a-hilaly (approved):
This is awesome! Thank you @nnbu @TiberiuGC !!

PR #225, opened 2024-09-27
Fix VPC optional DNS attribute handling

gfrey:
Issue #: aws-controllers-k8s/community#1826

Description of changes: Only sync optional fields, if actually configured. More details are in the [commit message](https://github.com/aws-controllers-k8s/ec2-controller/commit/3ebf6f911692ff11a8361bf6239ab67b8fc3652a).

### DIFF
--- a/pkg/resource/vpc/hooks.go
+++ b/pkg/resource/vpc/hooks.go
@@ -310,14 +310,18 @@ func (rm *resourceManager) customUpdateVPC(
 	}
 
 	if delta.DifferentAt("Spec.EnableDNSSupport") {
-		if err := rm.syncDNSSupportAttribute(ctx, desired); err != nil {
-			return nil, err
+		if desired.ko.Spec.EnableDNSSupport != nil {
+			if err := rm.syncDNSSupportAttribute(ctx, desired); err != nil {
+				return nil, err
+			}
 		}
 	}
 
 	if delta.DifferentAt("Spec.EnableDNSHostnames") {
-		if err := rm.syncDNSHostnamesAttribute(ctx, desired); err != nil {
-			return nil, err
+		if desired.ko.Spec.EnableDNSHostnames != nil {
+			if err := rm.syncDNSHostnamesAttribute(ctx, desired); err != nil {
+				return nil, err
+			}
 		}
 	}
 


gfrey:
I'm wondering whether this should be accompanied by other changes:
1. The tests by default [set the attributes](https://github.com/aws-controllers-k8s/ec2-controller/blob/main/test/e2e/resources/vpc.yaml). This would cover up any regression, when they are required to be set (as they are right now).
2. [Documentation ](https://aws-controllers-k8s.github.io/community/reference/ec2/v1alpha1/vpc/#spec) should maybe highlight these values have defaults.
If so, I would need some guidance on how to best do those changes.

a-hilaly:
Thanks @gfrey ! This looks good to me! QQ: what's the behaviour you're trying to eliminate here? i think the idea was to let people remove the DNS attributes by deleting the field.

gfrey:
> Thanks @gfrey ! This looks good to me! QQ: what's the behaviour you're trying to eliminate here? i think the idea was to let people remove the DNS attributes by deleting the field.

@a-hilaly  I was trying to create a minimal setup. From the documentation the only required field was the CIDR block, but it was failing because of this. Also the linked issue made me believe this is not the wanted behavior.

a-hilaly (approved):
LGTM, thank you @gfrey !!
/lgtm

a-hilaly:
@gfrey i have second thoughts on this. While it indeed gonna work. There is still a tiny problem: the controller will still detect a delta on those two fields, ideally what we can do to even avoid getting a delta is, having a customprecompare function, that takes into consideration default fields. For example we can do something similar to https://github.com/aws-controllers-k8s/efs-controller/blob/main/pkg/resource/file_system/hooks.go#L310-L318

gfrey:
> @gfrey i have second thoughts on this. While it indeed gonna work. There is still a tiny problem: the controller will still detect a delta on those two fields, ideally what we can do to even avoid getting a delta is, having a customprecompare function, that takes into consideration default fields. For example we can do something similar to https://github.com/aws-controllers-k8s/efs-controller/blob/main/pkg/resource/file_system/hooks.go#L310-L318

@a-hilaly Thanks for the pointer, will take a look tomorrow and prepare a second PR.

a-hilaly:
Thank you Gereon!! :)

PR #226, opened 2024-10-01
Improve the DNS attribute default handling

gfrey:
This improves #225.

Using a custom pre-compare function the defaults can be set accordingly. This is now added with the defaults set to what the VPC defaults are. The `disallowSecurityGroupDefaultRules` option is also handled here, as it also has the issue describe in aws-controllers-k8s/community#1826.


a-hilaly (approved):
/lgtm

PR #236, opened 2025-01-31
Issue 2090/route table order dependence

gfrey:
Fixes issue [#2090](https://github.com/aws-controllers-k8s/community/issues/2090)

Description of changes:
- Configure generator for RouteTable to ignore the Routes from the generated delta code.
- Extend the `customPreCompare` code to take care of the routes (computing the delta ignoring the order of routes).
- Added unit test for the custom code.
- Change the `syncRoutes` code to use the delta from the compare phase.

### DIFF
--- a/generator.yaml
+++ b/generator.yaml
@@ -535,6 +535,8 @@ resources:
       Routes:
         custom_field:
           list_of: CreateRouteInput
+        compare:
+          is_ignored: true
       RouteTableID:
         print:
           path: Status.RouteTableID
--- a/pkg/resource/route_table/hooks.go
+++ b/pkg/resource/route_table/hooks.go
@@ -48,6 +48,9 @@ func (rm *resourceManager) syncRoutes(
 	rlog := ackrtlog.FromContext(ctx)
 	exit := rlog.Trace("rm.syncRoutes")
 	defer func(err error) { exit(err) }(err)
+
+	// To determine the required updates to the route table, the routes to be
+	// added and deleted will be collected first.
 	toAdd := []*svcapitypes.CreateRouteInput{}
 	toDelete := []*svcapitypes.CreateRouteInput{}
 
@@ -65,18 +68,27 @@ func (rm *resourceManager) syncRoutes(
 	}
 
 	switch {
+	// If the route table is created all routes need to be added.
 	case delta == nil:
 		toAdd = removeLocalRoute(desired.ko.Spec.Routes)
+	// If there are changes to the routes in the delta ...
 	case delta.DifferentAt("Spec.Routes"):
+		// ... iterate over all the differences ...
 		for _, diff := range delta.Differences {
+			// ... and if the current one is regarding the routes ...
 			if diff.Path.Contains("Spec.Routes") {
+				// ... take the routes to add from the left side of the diff ...
 				toAdd = diff.A.([]*svcapitypes.CreateRouteInput)
+				// ... and the routes to delete from the right side (see the
+				// customPreCompare function for information on the diff
+				// structure).
 				toDelete = diff.B.([]*svcapitypes.CreateRouteInput)
 			}
 		}
 	default: // nothing to do
 	}
 
+	// Finally delete and add the routes that were collected.
 	for _, route := range toDelete {
 		rlog.Debug("deleting route from route table")
 		if err = rm.deleteRoute(ctx, latest, *route); err != nil {
@@ -201,6 +213,8 @@ var computeTagsDelta = tags.ComputeTagsDelta
 
 // customPreCompare ensures that default values of types are initialised and
 // server side defaults are excluded from the delta.
+// The left side (`A`) of any `Spec.Routes` diff contains the routes to add, the
+// right side (`B`) the routes that must be deleted.
 func customPreCompare(
 	delta *ackcompare.Delta,
 	a *resource,
--- a/pkg/resource/route_table/hooks.go
+++ b/pkg/resource/route_table/hooks.go
@@ -15,7 +15,6 @@ package route_table
 
 import (
 	"context"
-	"reflect"
 	"strings"
 
 	svcapitypes "github.com/aws-controllers-k8s/ec2-controller/apis/v1alpha1"
@@ -229,8 +228,9 @@ func getRoutesDifference(desired, latest []*svcapitypes.CreateRouteInput) (toAdd
 	for _, routeA := range desired {
 		found := false
 		for idx, routeB := range toDelete {
-			if found = reflect.DeepEqual(routeA, routeB); found {
+			if delta := compareCreateRouteInput(routeA, routeB); len(delta.Differences) == 0 {
 				toDelete = remove(toDelete, idx)
+				found = true
 				break
 			}
 		}
--- a/pkg/resource/route_table/hooks.go
+++ b/pkg/resource/route_table/hooks.go
@@ -213,8 +213,9 @@ var computeTagsDelta = tags.ComputeTagsDelta
 
 // customPreCompare ensures that default values of types are initialised and
 // server side defaults are excluded from the delta.
-// The left side (`A`) of any `Spec.Routes` diff contains the routes to add, the
-// right side (`B`) the routes that must be deleted.
+// The left side (`A`) of any `Spec.Routes` diff contains the routes that are
+// desired, but do not exist. Analogously, the right side (`B`) contains the
+// routes that exist, but are not desired.
 func customPreCompare(
 	delta *ackcompare.Delta,
 	a *resource,
@@ -223,8 +224,8 @@ func customPreCompare(
 	a.ko.Spec.Routes = removeLocalRoute(a.ko.Spec.Routes)
 	b.ko.Spec.Routes = removeLocalRoute(b.ko.Spec.Routes)
 
-	toAdd := []*svcapitypes.CreateRouteInput{}
-	toDelete := b.ko.Spec.DeepCopy().Routes
+	desired := []*svcapitypes.CreateRouteInput{}
+	latest := b.ko.Spec.DeepCopy().Routes
 
 	remove := func(s []*svcapitypes.CreateRouteInput, i int) []*svcapitypes.CreateRouteInput {
 		if i < len(s)-1 { // if not last element just copy the last element to where the removed element was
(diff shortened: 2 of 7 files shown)


pkg/resource/route_table/hooks.go
```diff
@@ -276,6 +207,37 @@ func customPreCompare(
 ) {
 	a.ko.Spec.Routes = removeLocalRoute(a.ko.Spec.Routes)
 	b.ko.Spec.Routes = removeLocalRoute(b.ko.Spec.Routes)
+
+	toAdd := []*svcapitypes.CreateRouteInput{}
+	toDelete := make([]*svcapitypes.CreateRouteInput, len(b.ko.Spec.Routes))
+	copy(toDelete, b.ko.Spec.Routes) // Copy of the routes from b, that will be reduced while iterating
```
michaelhtm, line 213:
```suggestion
	toDelete := b.ko.Spec.DeepCopy().Routes
```

pkg/resource/route_table/hooks.go
michaelhtm, line 1: nit: can we also add an e2e test addressing the issue?

michaelhtm (commented):
Hey @gfrey, thanks for the contribution..left a few comments.

generator.yaml
```diff
@@ -445,6 +445,8 @@ resources:
       Routes:
         custom_field:
           list_of: CreateRouteInput
+        compare:
```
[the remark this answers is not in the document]
gfrey: [I didn't include the generator output](https://github.com/aws-controllers-k8s/ec2-controller/pull/236#issuecomment-2627515465), as I was assuming this happens in the pipeline.
- I've rebased the code to latest main (that used v0.43.0 generator). This required a force-push.
- Run the used generator (PR #243 has the update to v0.43.1, but isn't merged yet).

Rebase was necessary, as otherwise I'd get the following error:
```
==== building ec2-controller ====
Copying common custom resource definitions into ec2
Building Kubernetes API objects for ec2
Error: cannot checkout tag: tag reference not found
```

pkg/resource/route_table/hooks.go
```diff
@@ -61,34 +63,17 @@ func (rm *resourceManager) syncRoutes(
...
-				// latest has a route that is not desired; therefore, delete
-				toDelete = append(toDelete, latestRoute)
+	switch {
+	case delta == nil:
+		toAdd = removeLocalRoute(desired.ko.Spec.Routes)
+	case delta.DifferentAt("Spec.Routes"):
+		for _, diff := range delta.Differences {
+			if diff.Path.Contains("Spec.Routes") {
+				toAdd = diff.A.([]*svcapitypes.CreateRouteInput)
+				toDelete = diff.B.([]*svcapitypes.CreateRouteInput)
```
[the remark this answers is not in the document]
gfrey: I tried to add some. Does it help?

pkg/resource/route_table/hooks.go
gfrey: I agree, the `DeepCopy` is easier to read, but two things:
- Then we'd have to also use `toAdd = append(toAdd, routeA.DeepCopy())`, as otherwise `toAdd` would still be a reference to the original struct from the spec.
- Creating a `DeepCopy` feels expensive (didn't measure though), but that might be "premature optimization" on my end.

pkg/resource/route_table/hooks.go
gfrey: To replicate my initial problem, I'd need to understand when order of routes changes in the AWS API. Otherwise, I wouldn't know how to test this.

gfrey:
The failing ec2-kind-e2e test seems to be unrelated to my changes, as it also came up in #241.

generator.yaml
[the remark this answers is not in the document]
michaelhtm: hey @gfrey, we should be done with releases for now, 
the pipeline does not run the generator as of now...
That error you see is caused by the migration to aws-sdk-go-v2...thanks for rebasing

pkg/resource/route_table/hooks.go
michaelhtm: to be honest you don't necessarily have to do a deep copy unless you're changing the object..also we already do deep copies throughout the controller already, so an extra one wouldn't necessarily affect the time complexity

pkg/resource/route_table/hooks.go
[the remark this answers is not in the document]
michaelhtm: Yes, Thank you :)

michaelhtm (commented):
Thanks for your patience @gfrey,
The flaky tests are due to eventual consistencies with tagging..we will have a fix for it in our next release, but this looks good to me for now

pkg/resource/route_table/hooks.go
```diff
@@ -270,13 +213,46 @@ var computeTagsDelta = tags.ComputeTagsDelta
...
+			}
+		}
+		if !found {
+			toAdd = append(toAdd, routeA)
+		}
+	}
+
+	if len(toAdd) > 0 || len(toDelete) > 0 {
+		delta.Add("Spec.Routes", toAdd, toDelete)
+	}
```
michaelhtm, line 255: one small issue i see here, delta does not hold what to add and what to delete, instead the A in delta stores the desired, and the B in delta stores the latest. This is to inform the user in the logs. Although I do like how you're using it here (saves an extra check for toAdd and toDelete during the update call), it will be inconsistent with the rest of the other deltas :(

michaelhtm (commented):
one more comment

pkg/resource/route_table/hooks.go
michaelhtm: here's how we do a precompare in iam: https://github.com/aws-controllers-k8s/iam-controller/blob/6828e1c2e8fbc0b8f8cc1886eee38ba43ad1df71/pkg/resource/role/hooks.go#L384-L395

and then we recalculate during update: https://github.com/aws-controllers-k8s/iam-controller/blob/6828e1c2e8fbc0b8f8cc1886eee38ba43ad1df71/pkg/resource/role/hooks.go#L434-L480

gfrey: @michaelhtm Is it? What we end up in the delta is the difference between the routes in the table, where
- A holds the routes that are desired, but don't exist, and
- B holds the routes that exist but are not desired.
I had named to those `toAdd` (aka desired but don't exist), and `toDelete` (aka exist but not desired). Renaming could make that more obvious.

The tags example you presented does this differently: it mentions all tags, not only the different ones. This is not my preference (makes finding the difference much harder), but if that is ACK idiomatic, I'll do it :).

generator.yaml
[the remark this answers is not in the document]
gfrey: I did a rebase onto the v0.43.2 generator (yesterday).

pkg/resource/route_table/hooks.go
gfrey: We discussed this and decided that it is better to stick with the current convention of the differences in a delta for complex types to give the entire thing, but only the changed parts.

pkg/resource/route_table/hooks.go
```diff
@@ -62,36 +62,19 @@ func (rm *resourceManager) syncRoutes(
...
+	var toAdd, toDelete []*svcapitypes.CreateRouteInput
+
+	switch {
+	// If the route table is created all routes need to be added.
+	case delta == nil:
+		toAdd = removeLocalRoute(desired.ko.Spec.Routes)
+	// If there are changes to the routes in the delta ...
+	case delta.DifferentAt("Spec.Routes"):
+		toAdd, toDelete = filterDifferentRoutes(desired.ko.Spec.Routes, latest.ko.Spec.Routes)
+	default: // nothing to do
```
michaelhtm, line 74: why are we adding if delta is nil?
```suggestion
	if delta.DifferentAt("Spec.Routes") {
	    toAdd, toDelete = filterDifferentRoutes(desired.ko.Spec.Routes, latest.ko.Spec.Routes)
        }
```

michaelhtm (commented):
Thank you for the revision @gfrey!!
left a few more comments

pkg/resource/route_table/hooks.go
gfrey: wow, good catch! I added `delta` to the signature, because I was using the differences for routes from it. Now that we recompute using `getRoutesDifference`, this is no longer necessary.

pkg/resource/route_table/hooks.go
```diff
@@ -270,13 +199,55 @@ var computeTagsDelta = tags.ComputeTagsDelta
...
+	desired, latest := filterDifferentRoutes(a.ko.Spec.Routes, b.ko.Spec.Routes)
+
+	if len(desired) > 0 || len(latest) > 0 {
+		delta.Add("Spec.Routes", a.ko.Spec.Routes, b.ko.Spec.Routes)
+	}
+}
+
+// filterDifferentRoutes compares the desired and latest routes. It returns the
+// routes that are different and must be added or deleted.
+func filterDifferentRoutes(desired, latest []*svcapitypes.CreateRouteInput) (toAdd, toDelete []*svcapitypes.CreateRouteInput) {
```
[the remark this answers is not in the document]
gfrey: done

pkg/resource/route_table/hooks.go
```diff
@@ -270,13 +191,55 @@ var computeTagsDelta = tags.ComputeTagsDelta
...
+	}
+
+	// Routes that are desired, but already exist in latest, can be ignored. The
+	// toDelete slice is a copy of latest and will be slowly modified so that at
+	// the end it only contains routes that exist in latest, but are not
+	// desired.
+	for _, routeA := range desired {
+		found := false
+		for idx, routeB := range toDelete {
+			if found = reflect.DeepEqual(routeA, routeB); found {
```
michaelhtm, line 232: One final concern from me, I'm a little uneasy about using DeepEqual. Especially after taking a closer look at the content of CreateRouteInput, there seem to be a lot of ResourceReference we use there (VPCPeeringConnectionRef). If this field is set in desired, in a lot of cases (highly likely this one as well) it will not be set in latest. There's an example of this happening in this github issue: https://github.com/aws-controllers-k8s/community/issues/2361#issuecomment-2689208675

A better approach to this is to have something like this.
What do you think @a-hilaly @rushmash91

michaelhtm (commented):
Hi @gfrey, I have one more comment, it should be a small fix, and also rebasing to latest changes (they shouldn't affect your PR much), thanks!

pkg/resource/route_table/hooks.go
gfrey: Okay, not sure where I got the "idea" with the DeepEqual (copy'n paste I assume). I understand and agree that is a bad idea, as the references from the `CreateRouteInput` would never be set in latest. Therefore, DeepEqual will always be false.
I'm not sure, but I suspect that your comment is missing a critical link for a solution?!

I've tried to fix it by using the `compareCreateRouteInput` method. This method doesn't look at the references at all. NOTE that my assumption is, that these references for desired are resolved and replaced with IDs way earlier in the process! If that doesn't hold true, then this is NOT the solution.

michaelhtm: Yes, this was the example: https://github.com/aws-controllers-k8s/cloudwatchlogs-controller/blob/6eef9f52db16fd2dddd4c5afd155337c0e7a7a6a/pkg/resource/log_group/hook.go#L246-L264

a-hilaly (approved):
Awesome! dankeshoen her @gfrey !!

gfrey:
@michaelhtm @a-hilaly Is there anything missing to get this merged? Anything I can help with?

PR #240, opened 2025-02-14
Add custom update support for `VPCEndpoint` resource to handle `ModifyVpcEndpoint` calls

rushmash91:
fixes https://github.com/aws-controllers-k8s/community/issues/2296

Description of changes:

Handle updates that require the `ModifyVpcEndpoint` API. Ensure that when fields such as `SubnetIDs`, `RouteTableIDs`, `PolicyDocument`, `PrivateDNSEnabled`, `SecurityGroupIDs`, `DNSOptions`, or `IPAddressType` change, the controller constructs and sends the corresponding Modify calls to AWS. 

### DIFF
--- a/pkg/resource/vpc_endpoint/hooks.go
+++ b/pkg/resource/vpc_endpoint/hooks.go
@@ -21,6 +21,7 @@ import (
 	ackrtlog "github.com/aws-controllers-k8s/runtime/pkg/runtime/log"
 
 	"github.com/aws-controllers-k8s/ec2-controller/pkg/tags"
+	"github.com/aws/aws-sdk-go-v2/aws"
 	svcsdk "github.com/aws/aws-sdk-go-v2/service/ec2"
 	svcsdktypes "github.com/aws/aws-sdk-go-v2/service/ec2/types"
 )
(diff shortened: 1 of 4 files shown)


pkg/resource/vpc_endpoint/hooks.go
```diff
@@ -81,6 +82,73 @@ func (rm *resourceManager) customUpdateVPCEndpoint(
 		}
 	}
```
a-hilaly, line 84: How about we remove the `needsModification` boolean, and instead prepend a block that exists if there are no changes:
```
if !delta.DifferentExcept("Spec.Tags") { return .. }
```

a-hilaly: example: https://github.com/aws-controllers-k8s/acm-controller/blob/2f43b106c9d3a7e0c7bc2ade18e429a8ab02868e/templates/hooks/certificate/sdk_update_pre_build_request.go.tpl#L10-L12

a-hilaly (approved):
Thanks @rushmash91 !
/lgtm

PR #245, opened 2025-02-24
Fix NetworkACL default rules duplication during creation

rushmash91:
Fixes https://github.com/aws-controllers-k8s/community/issues/2241

Description of changes:

Default Entries are already No-op fields fro the controller but were not filtered during creation(leading to `multiple entries with same number` errors) .

- Modified the NetworkACL resource creation logic to filter out default rules (rule number 32767) when creating entries
- Replaced the approach of appending entries to using a filtered list of entries
- Added a test case that verifies default rules are not duplicated when explicitly included in the creation spec

### DIFF
--- a/templates/hooks/network_acl/sdk_create_post_set_output.go.tpl
+++ b/templates/hooks/network_acl/sdk_create_post_set_output.go.tpl
@@ -11,10 +11,17 @@
 	}
 
 	if len(desired.ko.Spec.Entries) > 0 {
-		//desired rules are overwritten by NetworkACL's default rules
-		ko.Spec.Entries = append(ko.Spec.Entries, desired.ko.Spec.Entries...)
+		// Filter out default rules and only keep desired entries
+		filteredEntries := []*svcapitypes.NetworkACLEntry{}
+		for _, entry := range desired.ko.Spec.Entries {
+			if entry.RuleNumber != nil && *entry.RuleNumber == int64(DefaultRuleNumber) {
+				continue
+			}
+			filteredEntries = append(filteredEntries, entry)
+		}
+		ko.Spec.Entries = filteredEntries
 		copy := ko.DeepCopy()
 		if err := rm.createEntries(ctx, &resource{copy}); err != nil {
-			rlog.Debug("Error while syncing routes", err)
+			rlog.Debug("Error while syncing entries", err)
 		}
 	}
--- a/apis/v1alpha1/ack-generate-metadata.yaml
+++ b/apis/v1alpha1/ack-generate-metadata.yaml
@@ -1,7 +1,7 @@
 ack_generate_info:
-  build_date: "2025-02-20T17:55:49Z"
+  build_date: "2025-02-24T21:18:33Z"
   build_hash: a326346bd3a6973254d247c9ab2dc76790c36241
-  go_version: go1.24.0
+  go_version: go1.22.5
   version: v0.43.2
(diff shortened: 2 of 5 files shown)


templates/hooks/network_acl/sdk_create_post_set_output.go.tpl
```diff
@@ -11,10 +11,17 @@
...
+			if entry.RuleNumber != nil && *entry.RuleNumber == int64(DefaultRuleNumber) {
+				continue
+			}
+			filteredEntries = append(filteredEntries, entry)
+		}
+		ko.Spec.Entries = filteredEntries
 		copy := ko.DeepCopy()
 		if err := rm.createEntries(ctx, &resource{copy}); err != nil {
-			rlog.Debug("Error while syncing routes", err)
+			rlog.Debug("Error while syncing entries", err)
```
a-hilaly, line 25: nice catch 👀

templates/hooks/network_acl/sdk_create_post_set_output.go.tpl
```diff
@@ -11,10 +11,17 @@
 	}
 
 	if len(desired.ko.Spec.Entries) > 0 {
-		//desired rules are overwritten by NetworkACL's default rules
-		ko.Spec.Entries = append(ko.Spec.Entries, desired.ko.Spec.Entries...)
+		// Filter out default rules and only keep desired entries
+		filteredEntries := []*svcapitypes.NetworkACLEntry{}
+		for _, entry := range desired.ko.Spec.Entries {
+			if entry.RuleNumber != nil && *entry.RuleNumber == int64(DefaultRuleNumber) {
```
a-hilaly, line 17: qq: do we need to check more than the DefaultRuleNumber? maybe protocol? or is it unecessary?

a-hilaly (commented):
👍

templates/hooks/network_acl/sdk_create_post_set_output.go.tpl
rushmash91: Yes, isn't required here. There can not be user-defined rules with the same name, even if the other fields are different.

templates/hooks/network_acl/sdk_create_post_set_output.go.tpl
```diff
@@ -11,10 +11,17 @@
 	}
 
 	if len(desired.ko.Spec.Entries) > 0 {
-		//desired rules are overwritten by NetworkACL's default rules
-		ko.Spec.Entries = append(ko.Spec.Entries, desired.ko.Spec.Entries...)
+		// Filter out default rules and only keep desired entries
+		filteredEntries := []*svcapitypes.NetworkACLEntry{}
+		for _, entry := range desired.ko.Spec.Entries {
```
michaelhtm, line 16: Would it be best if we ignore the AWS injected default rules from the createRules output?

This would mean users no longer need to define the default rules and could be a better UX

michaelhtm (commented):
Great job on this @rushmash91 🔥
left one comment

templates/hooks/network_acl/sdk_create_post_set_output.go.tpl
rushmash91: Users did not have to define the default rules and that behavior still remains. This change is to handle certain GitOps scenarios where default rules might be defined anyway, which was leading to duplication errors.

PR #247, opened 2025-02-25
Fix NetworkACL default rules sync

rushmash91:
Fixes https://github.com/aws-controllers-k8s/community/issues/2241

Description of changes:
1. Filter out AWS default rules (rule #32767) from both desired and actual states during synchronization
2. Implement clear and documented behavior for default rules:
   - If default rules are explicitly defined in spec -> they remain in spec
   - If default rules are not defined in spec -> they are ignored during sync but preserved in AWS

AWS automatically creates two default DENY ALL rules (rule #32767) for each NetworkACL:
- One for ingress traffic
- One for egress traffic

These default rules would appear in the resource spec after creation even when not defined in the original manifest. This caused issues with GitOps tools like ArgoCD which would see this as a difference between desired and actual state, leading to continuous reconciliation attempts.

### DIFF
--- a/apis/v1alpha1/ack-generate-metadata.yaml
+++ b/apis/v1alpha1/ack-generate-metadata.yaml
@@ -1,7 +1,7 @@
 ack_generate_info:
-  build_date: "2025-02-20T17:55:49Z"
+  build_date: "2025-02-24T21:18:33Z"
   build_hash: a326346bd3a6973254d247c9ab2dc76790c36241
-  go_version: go1.24.0
+  go_version: go1.22.5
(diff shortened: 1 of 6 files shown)


michaelhtm:
👍 
/lgtm

gfrey:
Hey @rushmash91, thanks for fixing this! I think the default NACL entries should also be remove from the comparison. At least I see deltas, that only differ in default entries in latest.
I would suggest adding a `customPreCompare` and removing those at least from latest (should default entries be allowed in the desired configuration?)

PR #252, opened 2025-03-25
feat(NetworkACL): Filter Default Rules from Delta

rushmash91:
fixes https://github.com/aws-controllers-k8s/community/issues/2374

Description of changes:
- Add a precompare hook for Delta to filter out default rules from desired and latest objects
- Move create resource logic to `createEntries` 

### DIFF
--- a/pkg/resource/network_acl/hooks.go
+++ b/pkg/resource/network_acl/hooks.go
@@ -88,10 +88,10 @@ func (rm *resourceManager) requiredFieldsMissingForCreateNetworkAcl(
 
 func (rm *resourceManager) createEntries(
 	ctx context.Context,
-	r *resource,
+	desired *resource,
 ) error {
-	if r.ko.Spec.Entries != nil {
-		if err := rm.syncEntries(ctx, r, nil); err != nil {
+	if desired.ko.Spec.Entries != nil {
+		if err := rm.syncEntries(ctx, desired, nil); err != nil {
 			return err
 		}
 	}
@@ -351,7 +351,6 @@ func (rm *resourceManager) syncEntries(
 			toAdd = append(toAdd, desiredEntry)
 		}
 	}
-
 	if latest != nil {
 		// Filter out AWS default rules from latest entries before comparison
 		// to ensure consistent state management between desired and actual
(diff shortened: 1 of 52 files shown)


michaelhtm (approved):
👍

pkg/resource/network_acl/hooks.go
```diff
@@ -611,3 +602,26 @@ func updateTagSpecificationsInCreateRequest(r *resource,
...
+// Filter out AWS default rules from latest entries before comparison
+// to ensure consistent state management between desired and actual
+func customPreCompare(
+	delta *ackcompare.Delta,
+	a *resource,
+	b *resource,
+) {
+	a.ko.Spec.Entries = filterDefaultRules(a.ko.Spec.Entries)
+	b.ko.Spec.Entries = filterDefaultRules(b.ko.Spec.Entries)
+}
```
a-hilaly, line 615: Why do we filter rules both in preCompare and in syncEntries?

pkg/resource/network_acl/hooks.go
```diff
@@ -382,6 +368,11 @@ func (rm *resourceManager) syncEntries(
 		}
 	}
 
+	// During create latest is nil, just add the entries when sync is called via createEntries
+	if latest == nil {
+		toAdd = desired.ko.Spec.Entries
+	}
```
a-hilaly, line 374: why at this level of the code? also `toAdd` is getting elements in L344. Maybe we need an if/else to make sure we're not executing unecessary logic?

a-hilaly (commented):
👍

pkg/resource/network_acl/hooks.go
rushmash91: I'll just move out the creation from sync, don't think its a create idea.

pkg/resource/network_acl/hooks.go
```diff
@@ -88,12 +88,16 @@ func (rm *resourceManager) requiredFieldsMissingForCreateNetworkAcl(
...
-			return err
+	filteredEntries := filterDefaultRules(desired.ko.Spec.Entries)
+	if len(filteredEntries) > 0 {
+		for _, entry := range filteredEntries {
+			if err := rm.createEntry(ctx, desired, *entry); err != nil {
+				return err
+			}
 		}
+	
 	}
```
a-hilaly, line 101:
```suggestion
		for _, entry := range filteredEntries {
			if err := rm.createEntry(ctx, desired, *entry); err != nil {
				return err
			}
		}
```

PR #255, opened 2025-03-28
feat: new resource - LaunchTemplate

michaelhtm:
Issue [#1841](https://github.com/aws-controllers-k8s/community/issues/1841)

Description of changes:
Adding support for LaunchTemplate.


apis/v1alpha1/types.go
```diff
@@ -2939,7 +3206,8 @@ type LaunchTemplateENASrdUDPSpecification struct {
 //
 // Describes an elastic inference accelerator.
 type LaunchTemplateElasticInferenceAccelerator struct {
-	Type *string `json:"type_,omitempty"`
+	Count *int64  `json:"count,omitempty"`
+	Type  *string `json:"type_,omitempty"`
```
a-hilaly, line 3210:
```suggestion
	Type  *string `json:"type,omitempty"`
```

pkg/resource/launch_template/hooks.go
```diff
@@ -0,0 +1,854 @@
...
+	}
+
+	var resp *svcsdk.DescribeLaunchTemplateVersionsOutput
+	resp, err = rm.sdkapi.DescribeLaunchTemplateVersions(ctx, input)
+	rm.metrics.RecordAPICall("READ_MANY", "DescribeLaunchTemplateVersions", err)
+	if err != nil {
+		return err
+	}
+
+	for _, elem := range resp.LaunchTemplateVersions {
```
a-hilaly, line 55: i think function can be generated using custom templates and `sdk_file_end`

pkg/resource/launch_template/hooks.go
```diff
@@ -0,0 +1,854 @@
...
+) (err error) {
+	rlog := ackrtlog.FromContext(ctx)
+	exit := rlog.Trace("rm.findLaunchTemplateVersion")
+	defer func() {
+		exit(err)
+	}()
+
+	input, err := newListLaunchTemplateVersionRequestPayload(r)
+
+	if err != nil {
```
a-hilaly, line 44:
```suggestion
	input, err := newListLaunchTemplateVersionRequestPayload(r)
	if err != nil {
```

pkg/resource/launch_template/sdk.go
```diff
@@ -0,0 +1,2603 @@
...
+		return nil, ackerr.NotFound
+	}
+
+	input, err := rm.newListRequestPayload(r)
+	if err != nil {
+		return nil, err
+	}
+	var resp *svcsdk.DescribeLaunchTemplatesOutput
+	resp, err = rm.sdkapi.DescribeLaunchTemplates(ctx, input)
+	rm.metrics.RecordAPICall("READ_MANY", "DescribeLaunchTemplates", err)
```
a-hilaly, line 80: Do we want make the `DescribeLaunchVersion` here instead?

pkg/resource/launch_template/hooks.go
```diff
@@ -0,0 +1,854 @@
...
+	svcsdk "github.com/aws/aws-sdk-go-v2/service/ec2"
+	svcsdktypes "github.com/aws/aws-sdk-go-v2/service/ec2/types"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	svcapitypes "github.com/aws-controllers-k8s/ec2-controller/apis/v1alpha1"
+	"github.com/aws-controllers-k8s/ec2-controller/pkg/tags"
+)
+
+// sdkFind returns SDK-specific information about a supplied resource
+func (rm *resourceManager) findLaunchTemplateVersion(
```
a-hilaly, line 31: is it `findLaunchTemplateVersion`? or more `setLaunchLaunchTemplateAttributes`

pkg/resource/launch_template/hooks.go
```diff
@@ -0,0 +1,854 @@
...
+	"github.com/aws/aws-sdk-go-v2/aws"
+	svcsdk "github.com/aws/aws-sdk-go-v2/service/ec2"
+	svcsdktypes "github.com/aws/aws-sdk-go-v2/service/ec2/types"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	svcapitypes "github.com/aws-controllers-k8s/ec2-controller/apis/v1alpha1"
+	"github.com/aws-controllers-k8s/ec2-controller/pkg/tags"
+)
+
+// sdkFind returns SDK-specific information about a supplied resource
```
a-hilaly, line 30:
```suggestion
// findLaunchTemplateVersion returns SDK-specific information about a supplied resource
```

apis/v1alpha1/generator.yaml
```diff
@@ -485,6 +502,60 @@ resources:
...
+          path: Status.ID
+          name: ID
+      Tags:
+        from:
+          operation: CreateTags
+          path: Tags    
+      DefaultVersion:
+        from:
+          operation: ModifyLaunchTemplate
+          path: LaunchTemplate.DefaultVersionNumber
```
a-hilaly, line 558: Can we make this field a string instead? i like integer, but looks like the api takes a string, maybe cause they support non integer versions?

pkg/resource/launch_template/hooks.go
```diff
@@ -0,0 +1,854 @@
...
+
+	// TODO (michaelhtm) Not sure if we should check if
+	// the defaultVersion is greater than latestVersion
+	//
+	// My proposal would be to return a terminal error
+	// since the launchTemplate's latestVersion will never
+	// increase intil we make updates...˘\\/(ヅ)\/˘˘
+	// if *desired.ko.Spec.DefaultVersion > *latest.ko.Status.LatestVersion {
+	// 	return ackerr.NewTerminalError(fmt.Errorf("desired version number is ahead of the latest version"))
+	// }
```
a-hilaly, line 736: I like it.

pkg/resource/launch_template/hooks.go
```diff
@@ -0,0 +1,854 @@
...
+	r *resource,
+) (*svcsdk.ModifyLaunchTemplateInput, error) {
+	res := &svcsdk.ModifyLaunchTemplateInput{}
+
+	if r.ko.Spec.Name != nil {
+		res.LaunchTemplateName = r.ko.Spec.Name
+	}
+	if r.ko.Spec.DefaultVersion != nil {
+		res.DefaultVersion = aws.String(fmt.Sprintf("%d", *r.ko.Spec.DefaultVersion))
+	}
```
a-hilaly, line 815: what if the users sets, spec.DefaultVersion to nil, are we finding a delta, updating with no versions, for ever?

a-hilaly (commented):
Thanks @michaelhtm ! some comments inline

pkg/resource/launch_template/hooks.go
michaelhtm: You're right, we should probably require this field after the resource is found

apis/v1alpha1/generator.yaml
```diff
@@ -485,6 +502,66 @@ resources:
...
+          input_fields:
+            LaunchTemplateId: ID
+        DescribeLaunchTemplates:
+          input_fields:
+            LaunchTemplateIds: ID
+          output_fields:
+            LaunchTemplate.LatestVersionNumber: LatestVersion
+            LaunchTemplate.DefaultVersionNumber: DefaultVersion
+    fields:
+      Data: {}
```
michaelhtm, line 549: nit: remove

apis/v1alpha1/launch_template.go
```diff
@@ -0,0 +1,95 @@
...
+	CreatedBy *string `json:"createdBy,omitempty"`
+	// The ID of the launch template.
+	// +kubebuilder:validation:Optional
+	ID *string `json:"id,omitempty"`
+	// The version number of the latest version of the launch template.
+	// +kubebuilder:validation:Optional
+	LatestVersion *int64 `json:"latestVersion,omitempty"`
+	// The entity that manages the launch template.
+	// +kubebuilder:validation:Optional
+	Operator *OperatorResponse `json:"operator,omitempty"`
```
michaelhtm, line 71: I believe this field does not make sense for kubernetes..if another entity is managing this launchTemplate, we might encounter diffs whenever the entity makes changes

apis/v1alpha1/launch_template.go
```diff
@@ -0,0 +1,95 @@
...
+// LaunchTemplateSpec defines the desired state of LaunchTemplate.
+//
+// Describes a launch template.
+type LaunchTemplateSpec struct {
+
+	// The information for the launch template.
+	// +kubebuilder:validation:Required
+	Data *RequestLaunchTemplateData `json:"data"`
+	// The version number of the default version of the launch template.
+	DefaultVersion *int64 `json:"defaultVersion,omitempty"`
```
michaelhtm, line 32: Having DefaultVersion in the spec is risky, and a one way decision. We need to discuss this before merging..

Currently, when a launchTemplate is created, and later updated, the DefaultVersion will no longer be matching the data in the spec. This may not be ideal IMO.

If we put the defaultVersion in the status, we can start ignoring the latestVersion there and combine them both into one field called Version.
This Version will reflect the data in the spec, and with every update, we can call the APIs necessary to Create a new version with the latest data, ensure that version is the defaultVersion, delete the older version.

Opinions?

pkg/resource/launch_template/hooks.go
michaelhtm: putting setResourceForLaunchTemplateData there since that's the setter with the most LOC

michaelhtm (commented):
adding thoughts..

pkg/resource/launch_template/sdk.go
michaelhtm: i'm not opposed to it..

templates/hooks/launch_template/sdk_update_pre_build_request.go.tpl
```diff
@@ -0,0 +1,18 @@
...
+		); err != nil {
+			return nil, err
+		}
+	}
+	if delta.DifferentAt("Spec.DefaultVersion") {
+		err = rm.updateDefaultVersion(ctx, desired)
+		if err != nil {
+			return desired, err
+		}
+	}
```
a-hilaly, line 14: Should we make this call after updating the template instead? I'm thinking what if someone modifies both the template and default version at once.

test/e2e/tests/test_launch_template.py
```diff
@@ -0,0 +1,218 @@
...
+        launch_template = ec2_validator.get_launch_template(launch_template_id=resource_id)
+        default_version = launch_template['DefaultVersionNumber']
+        latest_version = launch_template['LatestVersionNumber']
+
+        # assert default version hasn't changed yet
+        assert default_version == 1
+        assert latest_version == 2
+
+        assert default_version == cr['spec']['defaultVersion']
+        assert latest_version == cr['status']['latestVersion']
```
a-hilaly, line 123: So even if the user doesn't specific a default version, we patch one back to the CR? do we want that behaviour?

test/e2e/tests/test_launch_template.py
```diff
@@ -0,0 +1,218 @@
...
+        assert k8s.wait_on_condition(ref, "ACK.ResourceSynced", "True", wait_periods=5)
+
+        cr = k8s.get_resource(ref)
+
+        launch_template = ec2_validator.get_launch_template(launch_template_id=resource_id)
+        default_version = launch_template['DefaultVersionNumber']
+        latest_version = launch_template['LatestVersionNumber']
+        
+        # latest_version was asserted to be 2 earlier
+        assert default_version == latest_version == 2
```
a-hilaly, line 152: python things!

a-hilaly (commented):
Thanks @michaelhtm ! i left a couple questions below

templates/hooks/launch_template/sdk_update_pre_build_request.go.tpl
michaelhtm: mmm that makes sense

test/e2e/tests/test_launch_template.py
michaelhtm: yes, at least after the create Operation...or would it make sense to put it in status?

apis/v1alpha1/launch_template.go
a-hilaly: shall we remove it?

templates/hooks/launch_template/sdk_update_pre_build_request.go.tpl
```diff
@@ -0,0 +1,19 @@
...
+			desired.ko.Spec.Tags, latest.ko.Spec.Tags,
+		); err != nil {
+			return nil, err
+		}
+	}
+	// We want to update the defaultVersion after we create the new
+	// version if needed.
+	if delta.DifferentAt("Spec.DefaultVersion") {
+		defer func() {
+			err = rm.updateDefaultVersion(ctx, desired)
```
a-hilaly, line 13: the error will only be logged here.

a-hilaly: How about doing something like, if diff in DefaultVersion and not in template data etc... make the update. And at the end, if diff in DefaultVersion...

apis/v1alpha1/launch_template.go
michaelhtm: it's not affecting anything significant since it's in the status..so idk

a-hilaly (approved):
👍

a-hilaly:
Mmmm ... the test failures seem unrelated to the changes

michaelhtm:
retest

a-hilaly (approved):

PR #257, opened 2025-04-02
feat: Recreate TransitGateway if resource has deleted status

michaelhtm:
Issue #, if available:

Description of changes:
Currently if a TransitGateway is deleted on accident by calling the API 
or console, it still dangles with a `deleted` state.

With this change, the controller will recreate the TransitGateway if the 
resource has `deleted` state

### DIFF
--- a/pkg/resource/nat_gateway/hooks.go
+++ b/pkg/resource/nat_gateway/hooks.go
@@ -24,6 +24,22 @@ import (
 	"github.com/aws-controllers-k8s/ec2-controller/pkg/tags"
 )
 
+func isResourceDeleted(r *resource) bool {
+	if r.ko.Status.State == nil {
+		return true
+	}
+	status := *r.ko.Status.State
+	return status == string(svcsdktypes.NatGatewayStateDeleted)
+}
+
+func isResourcePending(r *resource) bool {
+	if r.ko.Status.State == nil {
+		return false
+	}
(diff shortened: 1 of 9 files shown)


pkg/resource/nat_gateway/hooks.go
```diff
@@ -40,6 +56,7 @@ func (rm *resourceManager) customUpdateNATGateway(
 	// an error, then the update was successful and desired.Spec
 	// (now updated.Spec) reflects the latest resource state.
 	updated = rm.concreteResource(desired.DeepCopy())
+	updated.ko.Status = latest.ko.Status
```
a-hilaly, line 59: Why is this needed

michaelhtm: i don't want the latest.Status to be discarded..the state was never being patched when i tested it locally.
https://github.com/aws-controllers-k8s/runtime/blob/07dcd404a9a7bc3efe890791b910f9ec41712926/pkg/runtime/reconciler.go#L781

michaelhtm: we do need some preCompareDeltas to ensure we ignore everything except tags tho..since that's the only field we can update

a-hilaly (approved):
👍

rushmash91 (commented):
thanks @michaelhtm !
/lgtm

PR #261, opened 2025-04-14
add vpcid during adoption

rushmash91:
fixes https://github.com/aws-controllers-k8s/community/issues/2419

Description of changes:
add `read_many_post_set_output` hook to populate `vpcid` during adoption of VPC Peering Connection

### DIFF
--- a/templates/hooks/vpc_peering_connection/sdk_read_many_post_set_output.go.tpl
+++ b/templates/hooks/vpc_peering_connection/sdk_read_many_post_set_output.go.tpl
@@ -1,4 +1,8 @@
 
+	// This prevents reference resolution errors when adopting existing resources where these fields are not provided in the manifest.
+	if ko.Spec.VPCID == nil && ko.Status.RequesterVPCInfo != nil && ko.Status.RequesterVPCInfo.VPCID != nil {
+		ko.Spec.VPCID = ko.Status.RequesterVPCInfo.VPCID
+	}
 
 	if r.ko.Spec.AccepterPeeringConnectionOptions != nil {
 		f0 := &svcapitypes.PeeringConnectionOptionsRequest{}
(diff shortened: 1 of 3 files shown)


templates/hooks/vpc_peering_connection/sdk_read_many_post_set_output.go.tpl
```diff
@@ -1,4 +1,8 @@
 
+	// This prevents reference resolution errors when adopting existing resources where these fields are not provided in the manifest.
+	if ko.Spec.VPCID == nil && ko.Status.RequesterVPCInfo != nil && ko.Status.RequesterVPCInfo.VPCID != nil {
+		ko.Spec.VPCID = ko.Status.RequesterVPCInfo.VPCID
+	}
```
michaelhtm, line 5: should we return an error that makes sense as else?
It seems like describevpcpeeringconnection does not return requesterVPCInfo if the peering connection is not active...

michaelhtm (commented):
Thanks @rushmash91 👍 
can we also have a test for this?

templates/hooks/vpc_peering_connection/sdk_read_many_post_set_output.go.tpl
rushmash91: `requester-vpc-info.vpc-id `field is always returned in the `DescribeVpcPeeringConnections` API response, regardless of the peering connection's status. Even for deleted etc.

the `requester-vpc-info.cidrBlock` is not returned.

something like this:
```
  conditions:
  - lastTransitionTime: "2025-04-14T02:24:03Z"
    message: Resource synced successfully
    reason: ""
    status: "True"
    type: ACK.ResourceSynced
  requesterVPCInfo:
    ownerID: "xxxxx"
    region: us-west-2
    vpcID: vpc-0....
  status:
    code: deleted
```

michaelhtm (approved):
👍 
/lgtm

PR #273, opened 2025-07-08
Use ServiceID to match VPC Endpoint Service in ReadMany operation

knottnt:
Issue #, if available: [2547](https://github.com/aws-controllers-k8s/community/issues/2547)

Description of changes:
- Add ServiceId as ReadMany operation matching field
- Add custom required fields check to sdkFind
- Use DescribeVpcEndpointServiceConfigurations.ServiceIds to filter API request
- Add bootstrapped VPC Endpoint Service to e2e tests
- Add test for adopting VPC Endpoint Service

### DIFF
--- a/apis/v1alpha1/ack-generate-metadata.yaml
+++ b/apis/v1alpha1/ack-generate-metadata.yaml
@@ -1,13 +1,13 @@
 ack_generate_info:
-  build_date: "2025-07-22T22:13:44Z"
-  build_hash: b2dc0f44e0b08f041de14c3944a5cc005ba97c8f
+  build_date: "2025-07-25T17:47:22Z"
+  build_hash: 21224288556d986791b01b9caf69e2e20e0591ca
   go_version: go1.24.5
-  version: v0.50.0
+  version: v0.50.0-1-g2122428
 api_directory_checksum: d162a6e9df2d4861d6c01d42047402b51f341293
 api_version: v1alpha1
 aws_sdk_go_version: v1.32.6
 generator_config_info:
-  file_checksum: a3b7769e44651dbf7cdf955e76e47458b3436550
+  file_checksum: e3c9cd11651288747aece625d6883dfefa718864
   original_file_name: generator.yaml
 last_modification:
   reason: API generation
--- a/apis/v1alpha1/ack-generate-metadata.yaml
+++ b/apis/v1alpha1/ack-generate-metadata.yaml
@@ -7,7 +7,7 @@ api_directory_checksum: d162a6e9df2d4861d6c01d42047402b51f341293
 api_version: v1alpha1
 aws_sdk_go_version: v1.32.6
 generator_config_info:
-  file_checksum: e6a6ff840d55df735215ac04826122ebf89eb79a
+  file_checksum: a3b7769e44651dbf7cdf955e76e47458b3436550
   original_file_name: generator.yaml
 last_modification:
   reason: API generation
--- a/apis/v1alpha1/generator.yaml
+++ b/apis/v1alpha1/generator.yaml
@@ -254,12 +254,6 @@ operations:
     operation_type:
       - Delete
     resource_name: VpcEndpointServiceConfiguration
-  CreateVpcEndpointServiceConfiguration:
-    output_wrapper_field_path: ServiceConfiguration
-  DeleteVpcEndpointServiceConfigurations:
-    operation_type:
-      - Delete
-    resource_name: VpcEndpointServiceConfiguration
   CreateFlowLogs:
     operation_type:
     - Create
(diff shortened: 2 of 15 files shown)


pkg/resource/elastic_ip_address/sdk.go
```diff
@@ -100,10 +100,8 @@ func (rm *resourceManager) sdkFind(
 	found := false
 	for _, elem := range resp.Addresses {
 		if elem.AllocationId != nil {
-			if ko.Status.AllocationID != nil {
-				if *elem.AllocationId != *ko.Status.AllocationID {
-					continue
-				}
+			if ko.Status.AllocationID == nil || *elem.AllocationId != *ko.Status.AllocationID {
+				continue
```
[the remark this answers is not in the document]
knottnt: Note, this degrades the performance to that of a Spec matching field where no match exists.

apis/v1alpha1/generator.yaml
```diff
@@ -1068,7 +1062,10 @@ resources:
       when:
       - path: Status.ServiceState
         in:
-        - available
+        - Available
+    list_operation:
+      match_fields:
+        - ServiceId
```
michaelhtm, line 1068: should we declare ServiceID as primaryKey instead? This would ensure it is assigned to `DescribeVpcEndpointServiceConfigurationsInput`

test/e2e/service_bootstrap.py
```diff
@@ -37,7 +38,8 @@ def service_bootstrap() -> Resources:
         ),
         NetworkLoadBalancer=NetworkLoadBalancer("e2e-vpc-ep-service-test"),
         AdoptedVPC=VPC(name_prefix="e2e-adopted-vpc", num_public_subnet=1, num_private_subnet=0),
-        TestTransitGateway=TransitGateway()
+        TestTransitGateway=TransitGateway(),
+        AdoptedVpcEndpointService=VpcEndpointServiceConfiguration(name_prefix="e2e-adopted-vpc-es"),
```
michaelhtm, line 42: I have no issue with bootstrapping this..but should we instead create the resource using k8s?
Something like this: https://github.com/aws-controllers-k8s/kafka-controller/blob/main/test/e2e/tests/test_configuration.py#L66

michaelhtm: That's one case. Another one is, I don't see a case where we need to reuse VPCEndpointService in any other controller, so i don't see the need for it being bootstrapped. It's your call though

knottnt: I think I'm leaning towards leaving it as a bootstrappable resource for now since it matches the pattern already being used elsewhere in this controller. It also helps manage the sub-resources needed to create the VPC Endpoint Service.

templates/hooks/vpc_endpoint_service_configuration/sdk_read_many_pre_build_request.go.tpl
```diff
@@ -0,0 +1,3 @@
+    if r.ko.Status.ServiceID == nil {
+		return nil, ackerr.NotFound
+	}
```
michaelhtm, line 3: should we do something like this instead? https://github.com/aws-controllers-k8s/ec2-controller/blob/474f7aae6a22016458f127c4e275f2455b44c738/generator.yaml#L235-L236
To keep it consistent throughout the controller?

https://github.com/aws-controllers-k8s/ec2-controller/blob/474f7aae6a22016458f127c4e275f2455b44c738/pkg/resource/transit_gateway_vpc_attachment/hooks.go#L100-L102

I'm also okay with this how it is..

apis/v1alpha1/ack-generate-metadata.yaml
```diff
@@ -1,13 +1,13 @@
 ack_generate_info:
-  build_date: "2025-06-16T02:14:15Z"
-  build_hash: e675923dfc54d8b6e09730098c3e3e1056d3c1e9
-  go_version: go1.24.0
-  version: v0.48.0
+  build_date: "2025-07-08T05:50:37Z"
+  build_hash: 580a56e85fed07f60089ef2cae634548b523981b
+  go_version: go1.24.3
+  version: v0.48.0-2-g580a56e-dirty
```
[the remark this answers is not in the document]
knottnt: Ended up using hooks instead of this code-gen change.

pkg/resource/vpc_endpoint_service_configuration/sdk.go
```diff
@@ -224,7 +232,7 @@ func (rm *resourceManager) sdkFind(
 func (rm *resourceManager) requiredFieldsMissingFromReadManyInput(
 	r *resource,
 ) bool {
-	return false
+	return rm.checkForMissingRequiredFields(r)
```
michaelhtm, line 235: this seems to be hardcoded..i don't see the generator.yaml change for it

michaelhtm:
Nice
/lgtm

PR #296, opened 2025-11-03
fix: ignore errors when vpc does not have default security groups

michaelhtm:
Issue [#2661](https://github.com/aws-controllers-k8s/community/issues/2661)

Description of changes:
Currently the controller has been returning an error if the VPC
securityGroup is not found. With this change, we will ignore errors if
we can't find the security group

### DIFF
--- a/pkg/resource/vpc/hooks.go
+++ b/pkg/resource/vpc/hooks.go
@@ -15,7 +15,6 @@ package vpc
 
 import (
 	"context"
-	"fmt"
 
 	ackcompare "github.com/aws-controllers-k8s/runtime/pkg/compare"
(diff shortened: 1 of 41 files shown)


pkg/resource/vpc/hooks.go
```diff
@@ -560,7 +565,7 @@ func (rm *resourceManager) getDefaultSGId(
 	}
 
 	if len(resp.SecurityGroups) == 0 {
-		return nil, fmt.Errorf("default security group not found")
```
knottnt, line 563: nit: It might be worth logging the error message.

knottnt (commented):
@michaelhtm could you resolve the merge conflicts in this branch?

PR #321, opened 2026-04-07
feat: Add support for PrivateDnsPreference

matrus2:
Hi, 

I would like to expose new schema for VPC Endpoint CRD:
**API and CRD Schema Updates:**
* Added `privateDNSPreference` (string) and `privateDNSSpecifiedDomains` (array of strings) to the `DNSOptions` section in both the CRD and Helm chart schemas, enabling users to specify these options in their VPC Endpoint resources.

This is one of the functionality we miss currently. 
Please provide any guidance in case I missed something.


knottnt (commented):
Thanks @matrus2 changes to generator.yaml overall look good. However, VPCEndpoint does use a [custom update function](https://github.com/aws-controllers-k8s/ec2-controller/blob/987ffba8fbef44bef3d35728caa8f6c02d536282/pkg/resource/vpc_endpoint/hooks.go#L127) that will also need to be updated support the new fields.

matrus2:
@knottnt Thank you for looking at it. I have updated `hooks.go` as suggested.

pkg/resource/vpc_endpoint/hooks.go
knottnt, line 1: Testing the update path for these new fields I'm seeing a number of issues. @matrus2 were you able validate the update behavior locally?

1. ModifyVpcEndpoint doesn't appear to universally support `PrivateDnsPreference` and can return an error similar to the following when it isn't supported`api error UnknownParameter: The parameter PrivateDnsPreference is not recognized`. To resolve this we'll need to add `delta.DifferentAt(` guards to the individual DNSOptions fields. 

2. For `PrivateDnsPreference` the API accepts uppercase versions of the enum, but returns lowercase in `DescribeVpcEndpoints`. For example setting to `VERIFIED_DOMAINS_AND_SPECIFIED_DOMAINS` will be returned as `verified-domains-and-specified-domains`. We'll need to use a custom comparison in delta.go that can normalize these values. Note, the reverse may also need to handled.

knottnt: You can find more info on influencing the comparison logic in our docs.

[field-level config docs](https://aws-controllers-k8s.github.io/docs/contributing/code-generation/field-config#comparison-control)
[comparison hooks](https://aws-controllers-k8s.github.io/docs/contributing/code-generation/hooks#comparison-hook-points)

knottnt (commented):
@matrus2 sorry for the delay. Testing the update path it looks like we may need some additional logic to handle API behavior for ModifyVpcEndpoint. 

Regarding the `verify-code-gen` test failure. It looks like some file changes that are normally made by our code-generator tool are missing. To resolve that you should just need to rebase/merge the latest changes from upstream and re-build the controller with our code-generator. You can find more info on running our code-generator [here](https://aws-controllers-k8s.github.io/docs/contributing/building-controller).

pkg/resource/vpc_endpoint/hooks.go
matrus2: ## TL;DR

I've check and tested the point you raised and made two fixes to the `DnsOptions` motivated by live testing.

### ModifyVpcEndpoint

On a ServiceNetwork VPC endpoint, `ModifyVpcEndpoint` silently returns `Return: true` for any `DnsOptions` change while ignoring it — no error.

| Modify attempt | Response | Actual effect |
|---|---|---|
| `PrivateDnsPreference` → `ALL_DOMAINS` | `Return: true` | none |
| `PrivateDnsSpecifiedDomains` → real FQDNs | `Return: true` | none |
| `PrivateDnsEnabled` toggle (control) | `InvalidParameter` error | none |

So DnsOptions is effectively immutable on ServiceNetwork endpoints. It is addressed by:
```go
ackerrors.NewTerminalError(fmt.Errorf("DnsOptions.PrivateDnsPreference cannot be modified for ServiceNetwork endpoints"))
```

**Guard runs before `tags.Sync`.** Previously, a desired state touching both tags and immutable DNS options would sync tags first, then terminal-error — partially applying user intent. The guard now short-circuits before any side effects.

### Case-normalization concern — tested against a live endpoint:
Applied `privateDNSPreference: verified-domains-and-specified-domains` (lowercase-hyphen, the exact form you flagged). After reconcile:
- `kubectl get vpcendpoint ... -o jsonpath='{.spec.dnsOptions.privateDNSPreference}'` → `VERIFIED_DOMAINS_AND_SPECIFIED_DOMAINS`
- `aws ec2 describe-vpc-endpoints ... --query 'VpcEndpoints[0].DnsOptions.PrivateDnsPreference'` → `VERIFIED_DOMAINS_AND_SPECIFIED_DOMAINS`
Both sides converge on the canonical uppercase-underscore form via ACK's late-init, so the direct pointer-string comparison in `delta.go` doesn't produce a drift loop. Also confirmed with `all_domains` → `ALL_DOMAINS` earlier.

pkg/resource/vpc_endpoint/hooks.go
```diff
@@ -73,6 +76,34 @@ func (rm *resourceManager) customUpdateVPCEndpoint(
...
+		if latest.ko.Spec.VPCEndpointType != nil && *latest.ko.Spec.VPCEndpointType == "ServiceNetwork" {
+			desiredDNS := desired.ko.Spec.DNSOptions
+			latestDNS := latest.ko.Spec.DNSOptions
+
+			// If the user is trying to set or change DNS options when they shouldn't
+			if desiredDNS != nil {
+				// 1. Check PrivateDnsPreference
+				if desiredDNS.PrivateDNSPreference != nil {
+					if latestDNS == nil || latestDNS.PrivateDNSPreference == nil ||
+						*desiredDNS.PrivateDNSPreference != *latestDNS.PrivateDNSPreference {
```
knottnt: Could this re-use `delta.DifferentAt("Spec.DNSOptions.PrivateDNSPreference")`?

pkg/resource/vpc_endpoint/hooks.go
```diff
@@ -73,6 +76,34 @@ func (rm *resourceManager) customUpdateVPCEndpoint(
...
+						return latest, ackerrors.NewTerminalError(fmt.Errorf("DnsOptions.PrivateDnsPreference cannot be modified for ServiceNetwork endpoints"))
+					}
+				}
+
+				// 2. Check PrivateDnsSpecifiedDomains (since they work hand-in-hand)
+				if len(desiredDNS.PrivateDNSSpecifiedDomains) > 0 {
+					if latestDNS == nil || len(latestDNS.PrivateDNSSpecifiedDomains) == 0 ||
+						!stringSlicesEqual(desiredDNS.PrivateDNSSpecifiedDomains, latestDNS.PrivateDNSSpecifiedDomains) {
```
knottnt: Could this re-use `delta.DifferentAt("Spec.DNSOptions.PrivateDNSSpecifiedDomains")`?

pkg/resource/vpc_endpoint/hooks.go
```diff
@@ -73,6 +76,34 @@ func (rm *resourceManager) customUpdateVPCEndpoint(
 	// (now updated.Spec) reflects the latest resource state.
 	updated = rm.concreteResource(desired.DeepCopy())
 
+	// ServiceNetwork DnsOptions are immutable in AWS (silently no-op).
+	// Guard before tags.Sync to avoid partial-apply.
+	if delta.DifferentAt("Spec.DNSOptions") && desired.ko.Spec.DNSOptions != nil {
```
knottnt: Q: Should this also check `DNSOptions.DNSRecordIPType` it looks it may also perform a silent no-op when set.

knottnt (commented):
Thanks @matrus2 for live testing the ServiceNetwork behavior. Left a few comments on the guard clause.

pkg/resource/vpc_endpoint/hooks.go
matrus2: Added

pkg/resource/vpc_endpoint/hooks.go
matrus2: reused

pkg/resource/vpc_endpoint/hooks.go
matrus2: reused

matrus2 (commented):
@knottnt Thank you for quick reply and detailed review. I incorporated changes you requested plus one finding during testing:

The ServiceNetwork guard now ends with a catch-all `TerminalError("DnsOptions cannot be modified for ServiceNetwork endpoints")`. This turned out to be necessary in practice: if a user removes the whole `spec.dnsOptions` block, ACK's delta marks only the parent path as different (child paths report `false`), so none of the per-field checks trigger. Verified live — the catch-all correctly produces `ACK.Terminal` in that scenario.

Let me know if anything else needs tweaking!

knottnt (approved):
/lgtm

knottnt:
Thanks @matrus2! Changes look good to me.

knottnt:
@matrus2 this change has been released in [v1.17.0](https://github.com/aws-controllers-k8s/ec2-controller/releases/tag/v1.17.0) of the ec2-controller.

PR #325, opened 2026-04-27
fix(security_group): treat omitted GroupID in self-referencing rules as equal to self-id

### LINKED ISSUE #2822: [ec2-controller] SecurityGroup rules with self-reference recreated with each reconciliation cycle
**Describe the bug**
SecurityGroups with self-referencing rules are created correctly,
However, those rules are recreated with each reconciliation cycle as the rules have a diff in the groupID field, e.g.:

```json
{
  "level": "info",
  "ts": "2026-03-12T00:00:00.000Z",
  "logger": "ackrt",
  "msg": "desired resource state has changed",
  "kind": "SecurityGroup",
  "namespace": "ns",
  "name": "example-sg",
  "account": "123456789012",
  "role": "",
  "region": "us-east-1",
  "is_adopted": false,
  "generation": 3,
  "diff": [
    {
      "Path": {
        "Parts": [
          "Spec",
          "IngressRules"
        ]
      },
      "A": [
        {
          "fromPort": 443,
          "ipProtocol": "tcp",
          "ipRanges": [
            {
              "cidrIP": "10.0.0.0/20",
              "description": "Allow VPC-internal HTTPS traffic"
            },
            {
              "cidrIP": "192.168.0.0/16",
              "description": "Allow VPC-internal HTTPS traffic"
            }
          ],
          "toPort": 443,
          "userIDGroupPairs": [
            {
              "description": "Cluster API to node groups",
              "groupID": "sg-0a1b2c3d4e5f00001",
              "userID": "123456789012"
            } 
          ]
        },
        {
          "fromPort": 53,
          "ipProtocol": "tcp",
          "toPort": 53,
          "userIDGroupPairs": [
            {
              "description": "Node to node CoreDNS TCP",
              "userID": "123456789012"
            }
          ]
        },
        {
          "fromPort": 53,
          "ipProtocol": "udp",
          "toPort": 53,
          "userIDGroupPairs": [
            {
              "description": "Node to node CoreDNS UDP",
              "userID": "123456789012"
            }
          ]
        }
      ],
      "B": [
        {
          "fromPort": 53,
          "ipProtocol": "udp",
          "toPort": 53,
          "userIDGroupPairs": [
            {
              "description": "Node to node CoreDNS UDP",
              "groupID": "sg-0a1b2c3d4e5f00002",
              "userID": "123456789012"
            }
          ]
        },
        {
          "fromPort": 53,
          "ipProtocol": "tcp",
          "toPort": 53,
          "userIDGroupPairs": [
            {
              "description": "Node to node CoreDNS TCP",
              "groupID": "sg-0a1b2c3d4e5f00002",
              "userID": "123456789012"
            }
          ]
        },
        {
          "fromPort": 443,
          "ipProtocol": "tcp",
          "ipRanges": [
            {
              "cidrIP": "10.0.0.0/20",
              "description": "Allow VPC-internal HTTPS traffic (LoadBalancer IP target type)"
            },
            {
              "cidrIP": "192.168.0.0/16",
              "description": "Allow VPC-internal HTTPS traffic (LoadBalancer IP target type)"
            }
          ],
          "toPort": 443,
          "userIDGroupPairs": [
            {
              "description": "Cluster API to node groups",
              "groupID": "sg-0a1b2c3d4e5f00001",
              "userID": "123456789012"
            }
          ]
        }
      ]
    }
  ]
}
```

Yesterday, we had the problem that the controller failed between the [delete-create steps](https://github.com/aws-controllers-k8s/ec2-controller/blob/main/pkg/resource/security_group/hooks.go#L178-L183) which lead to missing rules in the SecurityGroup. As those were for EKS-cluster internal access to the coredns pods, network connectivity for the cluster workloads was degraded.

**Steps to reproduce**
Apply a SecurityGroup with self-referencing rules (by omitting the groupID field):

```yaml
kind: SecurityGroup
metadata:
  name: example-sg
  namespace: ns
spec:
  description: Test security group
  egressRules:
  - ipProtocol: "-1"
    ipRanges:
    - cidrIP: 0.0.0.0/0
      description: Allow all egress
  ingressRules:
  - fromPort: 443
    ipProtocol: tcp
    ipRanges:
    - cidrIP: 10.0.0.0/20
      description: Allow VPC-internal HTTPS traffic
    - cidrIP: 192.168.0.0/16
      description: Allow VPC-internal HTTPS traffic
    userIDGroupPairs:
    - description: Cluster API to node groups
      groupID: sg-0a1b2c3d4e5f00001
      userID: "123456789012"
  - fromPort: 53
    ipProtocol: tcp
    toPort: 53
    userIDGroupPairs:
    - description: Node to node CoreDNS TCP
      userID: "123456789012"
  - fromPort: 53
    ipProtocol: udp
    toPort: 53
    userIDGroupPairs:
    - description: Node to node CoreDNS UDP
      userID: "123456789012"
  name: example-sg
  vpcID: vpc-0a1b2c3d4e5f00001
```

Observe the changing Security group rule ID with each reconciliation cycle for the DNS rules

```bash
aws ec2 describe-security-group-rules \
  --filters Name=group-id,Values=sg-0a1b2c3d4e5f00001 \
  --region us-east-1 \
  --query 'SecurityGroupRules[].{RuleId:SecurityGroupRuleId,Direction:IsEgress,Proto:IpProtocol,From:FromPort,To:ToPort,CIDR:CidrIpv4,GroupId:ReferencedGroupInfo.GroupId}' \
  --output table
```

Initial rules

```
----------------------------------------------------------------------------------------------------------
|                                       DescribeSecurityGroupRules                                       |
+----------------+------------+-------+-----------------------+--------+-------------------------+-------+
|      CIDR      | Direction  | From  |        GroupId        | Proto  |         RuleId          |  To   |
+----------------+------------+-------+-----------------------+--------+-------------------------+-------+
|  10.0.0.0/20   |  False     |  443  |  None                 |  tcp   |  sgr-020b4a1388e3exxxx  |  443  |
|  None          |  False     |  53   |  sg-0a1b2c3d4e5f00002 |  tcp   |  sgr-007a0a79786b1xxxx  |  53   |
|  None          |  False     |  443  |  sg-0a1b2c3d4e5f00001 |  tcp   |  sgr-0678c73d6eca6xxxx  |  443  |
|  None          |  False     |  53   |  sg-0a1b2c3d4e5f00002 |  udp   |  sgr-09a273d6fba24xxxx  |  53   |
|  192.168.0.0/16|  False     |  443  |  None                 |  tcp   |  sgr-0556f86cd63a5xxxx  |  443  |
|  0.0.0.0/0     |  True      |  -1   |  None                 |  -1    |  sgr-06a7a8ce1891dxxxx  |  -1   |
+----------------+------------+-------+-----------------------+--------+-------------------------+-------+
```

After the next reconciliation

```
----------------------------------------------------------------------------------------------------------
|                                       DescribeSecurityGroupRules                                       |
+----------------+------------+-------+-----------------------+--------+-------------------------+-------+
|      CIDR      | Direction  | From  |        GroupId        | Proto  |         RuleId          |  To   |
+----------------+------------+-------+-----------------------+--------+-------------------------+-------+
|  10.0.0.0/20   |  False     |  443  |  None                 |  tcp   |  sgr-020b4a1388e3exxxx  |  443  |
|  None          |  False     |  53   |  sg-0a1b2c3d4e5f00002 |  tcp   |  sgr-00bb9cacbc638xxxx  |  53   |
|  None          |  False     |  443  |  sg-0a1b2c3d4e5f00001 |  tcp   |  sgr-0678c73d6eca6xxxx  |  443  |
|  None          |  False     |  53   |  sg-0a1b2c3d4e5f00002 |  udp   |  sgr-01904651746e7xxxx  |  53   |
|  192.168.0.0/16|  False     |  443  |  None                 |  tcp   |  sgr-0556f86cd63a5xxxx  |  443  |
|  0.0.0.0/0     |  True      |  -1   |  None                 |  -1    |  sgr-06a7a8ce1891dxxxx  |  -1   |
+----------------+------------+-------+-----------------------+--------+-------------------------+-------+
```

**Expected outcome**
Differences in the groupID for self-referencing rules do not trigger a recreation of the rule.

**Environment**

* Kubernetes version -> 1.34
* Using EKS (yes/no), if so version? -> yes, eks.14
* AWS service targeted (S3, RDS, etc.) -> EC2

kbujanecki-dt:
## Issue

Fixes [aws-controllers-k8s/community#2822](https://github.com/aws-controllers-k8s/community/issues/2822)

## Description

`SecurityGroup` rules whose `userIDGroupPairs[*].groupID` is omitted — the canonical way to express a self-reference, since the SG's own ID is unknown until AWS assigns it — are recreated on every reconciliation cycle.

Root cause is in `compareIPPermission` (`pkg/resource/security_group/sdk.go`), which compares `UserIDGroupPairs` with `equality.Semantic.Equalities.DeepEqual`. On `DescribeSecurityGroups` read-back AWS auto-fills **`GroupId`, `UserId`, and (optionally) `GroupName`** for self-referencing pairs, while the desired spec retains them as `nil`. The two sides therefore never compare equal → permanent diff → `RevokeSecurityGroupIngress` + `AuthorizeSecurityGroupIngress` on every reconcile.

This is the bug reported in #2822, with the additional operational severity that a controller crash *between* the Revoke and Authorize leaves the rule missing — degrading any workload that depends on the SG (CoreDNS node-to-node traffic in the reporter's reproducer).

## Fix

Add a small helper `normalizeSelfRefRules(r *resource) *resource` in `pkg/resource/security_group/hooks.go` and call it on **both** `desired` and `latest` at the top of `syncSGRules`.

Behaviour:

- **Returns a deep copy.** The caller's `desired` / `latest` are never mutated, so downstream code (status patching, late-initialisation, requeue) cannot accidentally lose user-set `GroupID`/`UserID` values.
- **Identifies a pair as a self-reference** when `GroupID` is `nil` *or* equals `r.ko.Status.ID`.
- **Clears the server-fillable fields** (`GroupID`, `UserID`, `GroupName`) for self-ref pairs only. `Description`, ports, protocol, IP/IPv6 ranges, prefix lists, cross-SG pairs, and IP-CIDR-only rules are untouched.

This makes the two sides semantically equivalent for the diff check while leaving the actual AWS API calls untouched: the existing outbound path (`createSecurityGroupRules` further down `hooks.go`) already canonicalises `nil` → self-id before `AuthorizeSecurityGroup{Ingress,Egress}`. The change is symmetric — the inbound side now matches the outbound convention.

The fix lives in the manually-maintained `hooks.go` rather than in the generated `compareIPPermission` in `sdk.go`, so it survives code-generator regenerations.

### Scope and follow-up

- Only **self-references** are addressed. Cross-SG / cross-account pairs that omit server-filled fields (e.g. `UserID`, `VPCID`) may still exhibit perpetual diffs; out of scope here.
- The runtime-level `Delta` computed *above* `customUpdateSecurityGroup` still observes the raw (unfilled vs filled) `Spec.IngressRules` each cycle and dispatches into `syncSGRules`. That dispatch is now harmless: the normalised diff produces empty `toAdd`/`toDelete` sets and **no AWS API call is made**. Eliminating that cosmetic log noise requires a normalisation hook above the generated update path — left for a follow-up.

### Unit tests

`pkg/resource/security_group/hooks_test.go`:

**`TestNormalizeSelfRefRules`** (10 subtests):

- `nil resource returns nil`
- `nil status ID returns a copy untouched`
- `does not mutate the input resource (deep-copy guarantee)`
- `strips server-fillable fields on self-ref ingress pair`
- `strips server-fillable fields on self-ref egress pair`
- `preserves cross-SG pair untouched`
- `nil GroupID counts as self-ref and clears UserID/GroupName`
- `mixed pair: only self stripped, cross-SG kept`
- `handles nil rule and nil pair entries`
- `preserves description on self-ref pair`

**`TestContainsRule_SelfRef_AfterNormalisation`** — locks the fix in at the comparison layer:

1. Asserts the pre-condition (`containsRule` returns `false` for canonical desired vs AWS-filled latest *without* normalisation — the bug exists),
2. Asserts the fix (`containsRule` returns `true` after `normalizeSelfRefRules` on both sides),
3. Asserts symmetry (both `desired→latest` and `latest→desired` directions match, so neither branch of `syncSGRules` schedules churn).

```
$ go test -v -run 'TestNormalizeSelfRefRules|TestContainsRule_SelfRef_AfterNormalisation' \
    ./pkg/resource/security_group/
PASS
ok  github.com/aws-controllers-k8s/ec2-controller/pkg/resource/security_group  0.589s
```

Full package test suite (`go test ./pkg/resource/security_group/...`) and module-wide `go build ./...` both pass.

### End-to-end against real AWS

Built the controller from this branch into a local image, deployed to k3d v1.34, applied the exact `SecurityGroup` CR pattern from #2822 (self-referencing ingress on TCP/53, `groupID` omitted, only `description` set) against a sandbox AWS account in `eu-central-1`. Observed across multiple reconcile cycles (requeue period 30s):

| Check | Pre-fix | Post-fix (this branch) |
|---|---|---|
| `Ready=True` time | <1s | <1s |
| `securityGroupRuleID` in `.status.rules` | changes every cycle | **stable** |
| `AuthorizeSecurityGroupIngress` API calls (post-create) | every cycle | **0** |
| `RevokeSecurityGroupIngress` API calls | every cycle | **0** |

Pre-fix log excerpt (rule ID changing between cycles 30s apart):

```
11:40:47  securityGroupRuleID: sgr-0261c49254bedc422
11:43:21  securityGroupRuleID: sgr-020e2ebec1bc13397
```

Post-fix log excerpt (same workload, multi-cycle window):

```
13:23:??  sgr-0d71004b23dcd7309
13:25:??  sgr-0d71004b23dcd7309   ← unchanged
$ kubectl logs ... | grep -E 'AuthorizeSecurityGroup|RevokeSecurityGroup'
(no output)
```

## Backwards compatibility

None. The helper operates on a deep copy used only for diff comparison — it does not alter what is sent to AWS. Non-self-referencing rules (cross-SG, IP/IPv6 CIDR, prefix-list) are untouched. Behaviour for existing CRs that already specified `groupID` explicitly to a different SG is unchanged.

## Files changed

- `pkg/resource/security_group/hooks.go` — adds `normalizeSelfRefRules` and rebinds `desired`/`latest` to its result at the top of `syncSGRules`.
- `pkg/resource/security_group/hooks_test.go` — new file with `TestNormalizeSelfRefRules` (10 subtests) and `TestContainsRule_SelfRef_AfterNormalisation`.


pkg/resource/security_group/hooks.go
michaelhtm, line 1: Can we also fix how we compare rules? We could have a custom compare, to avoid getting diffs

michaelhtm (commented):
Thanks @kbujanecki-dt 
I just have one comment
/test all

pkg/resource/security_group/hooks.go
kbujanecki-dt: @michaelhtm Good call - done in the latest push.

kbujanecki-dt: @michaelhtm Can you take another look at this?

generator.yaml
michaelhtm, line 1: looks like the generator.yaml in apis/v1alpha1 was not regenerated.

pkg/resource/security_group/hooks.go
```diff
@@ -129,6 +129,85 @@ func (rm *resourceManager) referencesResolved(
...
+// customPreCompare is injected at the top of the generated
+// newResourceDelta (see delta.go) via the `delta_pre_compare` hook in
+// generator.yaml. It canonicalises self-referencing UserIDGroupPairs on
+// both sides in place so that the subsequent field-by-field DeepEqual
+// does not report a spurious diff on Spec.IngressRules / Spec.EgressRules
+// when the only divergence is server-fill on self-references.
+//
+// Mutating a and b directly matches the convention used by RouteTable,
+// NetworkAcl, and VPC in this repo, and is safe here because:
+//   - ACK runtime patches only latest.ko.Status back to k8s, never Spec.
```
michaelhtm: this statement is not completely true..we patch Spec on Create output and update output. In the case of SecurityGroup it should be okay, since sdkUpdate returns deepCopy of desired. This comment may be misleading tho

pkg/resource/security_group/hooks_test.go
michaelhtm, line 1: can we also add an e2e test for this?

michaelhtm (commented):
Thanks @kbujanecki-dt 
just had a few more comments

pkg/resource/security_group/hooks_test.go
kbujanecki-dt: Test added. Also updated the missing generated.yaml and regenerated the APIs (generator.yaml and ack-generate-metadata.yaml). pls review again.

shabbskagalwala:
@michaelhtm @kbujanecki-dt this is what we ran into as well. Would love to have this released so as to not cause outages in production environments

gustavodiaz7722:
Closing this pull request since fix has been applied in https://github.com/aws-controllers-k8s/ec2-controller/pull/359

PR #330, opened 2026-05-08
fix(vpc): Delete default SG rules at creation when disallowSecurityGroupDefaultRules is true

sapphirew:
Issue #, if available:

Description of changes:

Previously, the controller only deleted default security group rules during the update path (customUpdate):
https://github.com/aws-controllers-k8s/ec2-controller/blob/main/pkg/resource/vpc/hooks.go#L351, which requires a delta between desired and latest. When a VPC is created with disallowSecurityGroupDefaultRules: true, the reconciler runs the create path and then goes directly to late initialization — it never enters the update path in the same reconcile. The resource is then marked synced and requeued for 10h, so no second reconcile happens before the test asserts.

Now the `sdk_create_post_set_output` hook checks the desired spec and deletes the default SG rules immediately after creation if the user requested it, ensuring the test_disallow_default_security_group_rule e2e test passes consistently.

### DIFF
--- a/templates/hooks/vpc/sdk_create_post_set_output.go.tpl
+++ b/templates/hooks/vpc/sdk_create_post_set_output.go.tpl
@@ -12,6 +12,15 @@
 	sgDefaultRulesExist, err := rm.hasSecurityGroupDefaultRules(ctx, &resource{ko})
 	if err != nil {
 		return nil, err
-	} else {
-		ko.Status.SecurityGroupDefaultRulesExist = &sgDefaultRulesExist
+	}
+	ko.Status.SecurityGroupDefaultRulesExist = &sgDefaultRulesExist
+
+	// If the user requested disallowing default SG rules, requeue so the
+	// update path can handle the deletion. We avoid making the delete call
+	// here because the runtime treats sdkCreate failures as resource
+	// creation failures, which could orphan the VPC.
+	if sgDefaultRulesExist && desired.ko.Spec.DisallowSecurityGroupDefaultRules != nil && *desired.ko.Spec.DisallowSecurityGroupDefaultRules {
+		ackcondition.SetSynced(&resource{ko}, corev1.ConditionFalse, aws.String("VPC created, requeue to delete default security group rules"), nil)
+		err = ackrequeue.NeededAfter(fmt.Errorf("VPC created but default security group rules need to be deleted"), time.Second)
+		return &resource{ko}, err
 	}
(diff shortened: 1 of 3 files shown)


config/controller/kustomization.yaml
```diff
@@ -6,4 +6,4 @@ kind: Kustomization
 images:
 - name: controller
   newName: public.ecr.aws/aws-controllers-k8s/ec2-controller
-  newTag: 1.12.0
+  newTag: 1.11.1
```
knottnt: Make sure you've pulled in the latest upstream tags. This version change is unnecessary.

sapphirew: whoops, fixed

templates/hooks/vpc/sdk_create_post_set_output.go.tpl
```diff
@@ -12,6 +12,14 @@
 	sgDefaultRulesExist, err := rm.hasSecurityGroupDefaultRules(ctx, &resource{ko})
 	if err != nil {
 		return nil, err
-	} else {
-		ko.Status.SecurityGroupDefaultRulesExist = &sgDefaultRulesExist
 	}
+	// If the user requested disallowing default SG rules at creation time,
+	// delete them now before returning the created resource.
+	if sgDefaultRulesExist && desired.ko.Spec.DisallowSecurityGroupDefaultRules != nil && *desired.ko.Spec.DisallowSecurityGroupDefaultRules {
+		if err = rm.deleteSecurityGroupDefaultRules(ctx, &resource{ko}); err != nil {
```
knottnt: Generally, we want to avoid making additional api calls in sdkCreate. The ACK runtime interprets AWS API failures in this function as a failure to create the AWS resource. If this API call fails the controller could end up orphaning the created VPC.

knottnt (commented):
@sapphirew I'm not totally sure I understand the diagnosis of the post-create read call overwriting the `spec.disallowSecurityGroupDefaultRules` field in the desired manifest. That Read [call](https://github.com/aws-controllers-k8s/runtime/blob/a4b7540ebdfc78f056e0447c08c02479c751b58c/pkg/runtime/reconciler.go#L812) is only used to set the status fields. Looking at the test I wonder if the issue is that we need to force synced to false in sdkCreate to ensure that the controller reconciles the resource's update only operations.

templates/hooks/vpc/sdk_create_post_set_output.go.tpl
sapphirew: Fair enough, updated to set `ACK.ResourceSynced = False` to trigger a quick re-reconcile in the update path, instead of making additional api calls in sdkCreate

knottnt (approved):
/lgtm

PR #331, opened 2026-05-11
fix(test): Fix flaky VPC e2e tests for SG rule deletion

sapphirew:
Issue #, if available:

Description of changes:

The test_crud_tags test failed because the ACK runtime's EnsureTags persist ACK system tags (services.k8s.aws/controller-version, services.k8s.aws/namespace) into spec.tags during reconciliation. Filter out tags with the `services.k8s.aws/` prefix before asserting user tag counts and values.
    
The `test_disallow_default_security_group_rule` test failed because the controller was too busy processing other resources to complete the reconcile within the fixed 10-second sleep. Replace `time.sleep` with `wait_on_condition("ACK.ResourceSynced", "True")` which polls until the controller signals reconciliation is complete.

### DIFF
--- a/test/e2e/tests/test_vpc.py
+++ b/test/e2e/tests/test_vpc.py
@@ -249,7 +249,9 @@ class TestVpc:
         (ref, cr) = simple_vpc
         resource_id = cr["status"]["vpcID"]
 
-        time.sleep(CREATE_WAIT_AFTER_SECONDS)
+        # Wait for the controller to finish reconciling (SG rule deletion
+        # happens inline during creation but the controller may be busy)
+        assert k8s.wait_on_condition(ref, "ACK.ResourceSynced", "True", wait_periods=5)
 
         # Check VPC exists in AWS
         ec2_validator = EC2Validator(ec2_client)


test/e2e/tests/test_vpc.py
```diff
@@ -203,11 +204,12 @@ def test_crud_tags(self, ec2_client, simple_vpc):
             actual=vpc["Tags"],
         )
                
-        # Only user tags should be present in Spec
+        # Only user tags should be present in Spec (filter out ACK system tags)
         resource = k8s.get_resource(ref)
-        assert len(resource["spec"]["tags"]) == 1
-        assert resource["spec"]["tags"][0]["key"] == "updatedtagkey"
-        assert resource["spec"]["tags"][0]["value"] == "updatedtagvalue"
+        user_spec_tags = [t for t in resource["spec"]["tags"] if not t["key"].startswith("services.k8s.aws/")]
```
gustavodiaz7722: Can we update this to use shared util functions for asserting without ack tags? Heres an example from emr-serverless 

https://github.com/aws-controllers-k8s/emrserverless-controller/blob/main/test/e2e/tests/test_application.py#L193

```
        updated_tags = {"environment": "production", "team": "data-platform"}
        cr = k8s.get_resource(ref)
        cr["spec"]["tags"] = updated_tags
        k8s.replace_custom_resource(ref, cr)
        time.sleep(MODIFY_WAIT_AFTER_SECONDS)
        
        assert k8s.wait_on_condition(ref, "ACK.ResourceSynced", "True", wait_periods=10)
        
        response = emrserverless_client.list_tags_for_resource(resourceArn=application_arn)
        latest_tags = response["tags"]
        
        tags.assert_ack_system_tags(tags=latest_tags)
        tags.assert_equal_without_ack_tags(expected=updated_tags, actual=latest_tags)
```

test/e2e/tests/test_vpc.py
```diff
@@ -166,10 +166,11 @@ def test_crud_tags(self, ec2_client, simple_vpc):
             actual=vpc["Tags"],
         )
         
-        # Only user tags should be present in Spec
-        assert len(resource["spec"]["tags"]) == 1
-        assert resource["spec"]["tags"][0]["key"] == "initialtagkey"
-        assert resource["spec"]["tags"][0]["value"] == "initialtagvalue"
+        # Only user tags should be present in Spec (filter out ACK system tags)
+        user_spec_tags = [t for t in resource["spec"]["tags"] if not t["key"].startswith("services.k8s.aws/")]
```
gustavodiaz7722: Same thing here, use shared utility functions for asserting without ack tags

test/e2e/tests/test_vpc.py
```diff
@@ -230,9 +232,10 @@ def test_crud_tags(self, ec2_client, simple_vpc):
             actual=vpc["Tags"],
         )
         
-        # Check user tags are removed from Spec
+        # Check user tags are removed from Spec (filter out ACK system tags)
```
gustavodiaz7722: Same thing here, use shared utility functions for asserting without ack tags

knottnt:
@sapphirew I think this is actually a bug in the ec2-controller. The system tags should be filtered before patching the spec.tags field.

test/e2e/tests/test_vpc.py
sapphirew: Good call, updated to use the shared lib for tags assertion

sapphirew: Looks like the tests already make tags assertion above. Going to drop this change until have a clear root cause.

sapphirew:
> @sapphirew I think this is actually a bug in the ec2-controller. The system tags should be filtered before patching the spec.tags field.

@knottnt I couldn't root cause the bug for tags in the ec2-controller yet. Going to reduce the scope of the fix to  `test_disallow_default_security_group_rule` only and follow up on that

michaelhtm:
@sapphirew 
System tags are not supposed to be patched to the resource. I'll continue investigating this bug and address it
/hold

michaelhtm:
This PR should fix system tags issues: https://github.com/aws-controllers-k8s/code-generator/pull/702

test/e2e/tests/test_vpc.py
```diff
@@ -249,7 +249,9 @@ def test_disallow_default_security_group_rule(self, ec2_client, simple_vpc):
         (ref, cr) = simple_vpc
         resource_id = cr["status"]["vpcID"]
 
-        time.sleep(CREATE_WAIT_AFTER_SECONDS)
+        # Wait for the controller to finish reconciling (SG rule deletion
+        # happens inline during creation but the controller may be busy)
```
knottnt, line 253: nit: I don't think this is true anymore after you previous [PR](https://github.com/aws-controllers-k8s/ec2-controller/pull/330/changes).
```suggestion
        # Wait for the controller to finish reconciling (SG rule deletion
        # happens in update path after initial creation)
```

PR #332, opened 2026-05-17
Fix EgressOnlyInternetGateway sdkFind returning wrong resource when Status.ID is nil

Mais316:
**Issue**

When multiple `EgressOnlyInternetGateway` (EIGW) resources exist across different VPCs in the same AWS account, creating a new EIGW fails with a terminal `Resource already exists` error — even though no EIGW exists on the target VPC.

The same class of bug also affects `FlowLog`: when `Status.FlowLogID` is `nil`, `DescribeFlowLogs` is called without a `FlowLogIds` filter, returning all flow logs in the account and potentially causing a false "already exists" match.

**Root Cause**

`requiredFieldsMissingFromReadManyInput` for `EgressOnlyInternetGateway` and `FlowLog` always returned `false`. When the resource's Status ID is `nil` (i.e. during initial creation before the resource is created in AWS), `newListRequestPayload` builds a Describe request with no ID filter. The API returns **all** resources in the account. `sdkFind` then takes the first result, incorrectly matches it as the desired resource, and returns `AlreadyExists`.

This is the same class of bug that was previously fixed for `ManagedPrefixList` and `LaunchTemplate`.

**Fix**

Regenerated the controller with an updated code generator that now correctly emits nil checks for renamed fields:

- **`pkg/resource/egress_only_internet_gateway/sdk.go`**: `requiredFieldsMissingFromReadManyInput` now returns `true` when `Status.ID` is `nil`, causing `sdkFind` to correctly return `NotFound` instead of issuing an unfiltered Describe call.
- **`pkg/resource/flow_log/sdk.go`**: `requiredFieldsMissingFromReadManyInput` now returns `true` when `Status.FlowLogID` is `nil`, and `newListRequestPayload` passes `FlowLogIds` to `DescribeFlowLogs` when the ID is set.

**Tests**

- **E2e test** (`test_egress_only_internet_gateway.py`): `test_multiple_eigws_different_vpcs` — creates two VPCs each with their own EIGW in the same account. Validates both EIGWs sync successfully with correct VPC attachments. Before the fix, the second EIGW would fail with `Resource already exists`.

**By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.**


Mais316:
@michaelhtm Kindly review when you have time as we hit a bug mentioned in description 🙏

lucky5star:
interesting approach.
was this implementation chosen primarily because of controller runtime reconciliation semantics, or were there limitations in the EC2 API responses that made this pattern more reliable?

devpulse-star:
hey sir 
thank u for ur PR
I was reading through this PR to better understand the reconcile flow in ACK controllers.
one thing I was curious about that.
if the EC2 resource changes outside Kubernetes between reconcile cycles, would this still keep the resource state in sync correctly?
still learning this part of the architecture, so I wonna ask.

michaelhtm:
Thanks @Mais316 
/test all

michaelhtm:
>interesting approach.
was this implementation chosen primarily because of controller runtime reconciliation semantics, or were there limitations in the EC2 API responses that made this pattern more reliable?

@lucky5star 
Using a `checkMissingIdentifier` is a pattern used across all controllers that store their Identifier in the resource status. The EC2 API does not accept vpcID in the describe input, so we mainly just ensure whether or not ACK has created the resource before.

michaelhtm:
>if the EC2 resource changes outside Kubernetes between reconcile cycles, would this still keep the resource state in sync correctly?

@devpulse-star 
Yes, there is a configurable resync period that ensures the controller checks for drift between the kubernetes resource and the AWS resource.

michaelhtm:
@Mais316 
the code-gen failure is mainly due to a newer version of code-generator. 
We will release a new version soon, can you regenerate the controller with it when you get a chance?
https://github.com/aws-controllers-k8s/code-generator/pull/704

Mais316:
> @Mais316 the code-gen failure is mainly due to a newer version of code-generator. We will release a new version soon, can you regenerate the controller with it when you get a chance? [aws-controllers-k8s/code-generator#704](https://github.com/aws-controllers-k8s/code-generator/pull/704)

Done :)

michaelhtm:
Hey @Mais316 
I just fixed this issue in code-generator. Once you re-run it now, it should generate the `checkRequiredFields` function, without needing the custom hook. 
https://github.com/aws-controllers-k8s/code-generator/pull/705
Can you try re-generating with code-generator upstream main?

Mais316:
> Hey @Mais316 I just fixed this issue in code-generator. Once you re-run it now, it should generate the `checkRequiredFields` function, without needing the custom hook. [aws-controllers-k8s/code-generator#705](https://github.com/aws-controllers-k8s/code-generator/pull/705) Can you try re-generating with code-generator upstream main?

@michaelhtm  looks working, please check when you have time.

michaelhtm:
Thanks @Mais316 
can you also rebase this repo? there seems to be a conflict

Mais316:
@michaelhtm kindly check when u have time

michaelhtm:
Thanks again @Mais316 
/ok-to-test
/lgtm

michaelhtm:
One quick thing, can you also update the PR description?

Mais316:
> One quick thing, can you also update the PR description?

done :)

PR #336, opened 2026-05-26
fix: prevent false drift in VPCEndpoint and SecurityGroup resources

### LINKED ISSUE #2695: [RDS controller]: constant loop of changing the db in the cloud
We are using ACK Controllers to manage our database resources, and we are observing a persistent issue: after a `DBinstance` is created, it regularly falls back into the configuring cycle.
In the controller logs we constantly see messages like:

```
{"level":"info","ts":"2025-11-17T08:05:49.709Z","logger":"ackrt","msg":"desired resource state has changed","kind":"DBInstance","namespace":"infra-production","name":"database-2025-11-09","account":"966321756598","role":"arn:aws:iam::966321756598:role/ack-ec2-controller-tooling","region":"eu-west-2","is_adopted":false,"generation":2,"diff":[{"Path":{"Parts":["Spec","AutoMinorVersionUpgrade"]},"A":null,"B":true},{"Path":{"Parts":["Spec","CACertificateIdentifier"]},"A":null,"B":"rds-ca-rsa2048-g1"},{"Path":{"Parts":["Spec","LicenseModel"]},"A":null,"B":"general-public-license"},{"Path":{"Parts":["Spec","MultiAZ"]},"A":null,"B":false},{"Path":{"Parts":["Spec","PreferredBackupWindow"]},"A":null,"B":"00:56-01:26"},{"Path":{"Parts":["Spec","PreferredMaintenanceWindow"]},"A":null,"B":"sat:23:48-sun:00:18"},{"Path":{"Parts":["Spec","StorageEncrypted"]},"A":null,"B":false},{"Path":{"Parts":["Spec","StorageThroughput"]},"A":null,"B":0},{"Path":{"Parts":["Spec","StorageType"]},"A":null,"B":"gp2"}]}
```

This behavior clearly doesn’t seem normal. While the controller repeatedly attempts to “configure” the instance, the database may become unavailable — which is unacceptable in production.
It would be very helpful to have guidance or a built-in mechanism to prevent this repeating reconfiguration loop.
Additionally, we see a related issue when `AutoMinorVersionUpgrade` is enabled:
the cloud database automatically upgrades to a newer minor version, while the version in the ACK manifest remains unchanged. After that, any further changes applied via the manifest get stuck, because the version mismatch prevents updates from proceeding.
Could you please clarify whether this is expected behavior, and whether there are recommended practices to avoid these situations?

michaelhtm:
Description of changes:
VPCEndpoint: Add is_iam_policy for PolicyDocument (semantic JSON compare
instead of string compare), add late_initialize for ServiceRegion.

SecurityGroup: Replace order-sensitive DeepEqual for IngressRules/
EgressRules with set-based containsRule comparison via delta_pre_compare
hook, preventing false diffs from rule ordering differences.

Resolves aws-controllers-k8s/community#2695

### DIFF
--- a/pkg/resource/security_group/delta.go
+++ b/pkg/resource/security_group/delta.go
@@ -41,26 +41,32 @@ func newResourceDelta(
 		delta.Add("", a, b)
 		return delta
 	}
+	if len(a.ko.Spec.IngressRules) != len(b.ko.Spec.IngressRules) {
+		delta.Add("Spec.IngressRules", a.ko.Spec.IngressRules, b.ko.Spec.IngressRules)
+	} else {
+		for _, aRule := range a.ko.Spec.IngressRules {
+			if !containsRule(b.ko.Spec.IngressRules, aRule) {
+				delta.Add("Spec.IngressRules", a.ko.Spec.IngressRules, b.ko.Spec.IngressRules)
+				break
+			}
+		}
+	}
+	if len(a.ko.Spec.EgressRules) != len(b.ko.Spec.EgressRules) {
+		delta.Add("Spec.EgressRules", a.ko.Spec.EgressRules, b.ko.Spec.EgressRules)
+	} else {
+		for _, aRule := range a.ko.Spec.EgressRules {
+			if !containsRule(b.ko.Spec.EgressRules, aRule) {
+				delta.Add("Spec.EgressRules", a.ko.Spec.EgressRules, b.ko.Spec.EgressRules)
+				break
+			}
+		}
+	}
 
 	if ackcompare.HasNilDifference(a.ko.Spec.Description, b.ko.Spec.Description) {
 		delta.Add("Spec.Description", a.ko.Spec.Description, b.ko.Spec.Description)
 	} else if a.ko.Spec.Description != nil && b.ko.Spec.Description != nil {
 		if *a.ko.Spec.Description != *b.ko.Spec.Description {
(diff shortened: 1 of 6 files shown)


knottnt:
Looks like PolicyDocument fix to use is_iam_policy was applied in https://github.com/aws-controllers-k8s/ec2-controller/pull/343

knottnt (approved):
@michaelhtm lgtm. Just need to resolve the merge conflicts.

pkg/resource/security_group/delta.go
```diff
@@ -41,26 +41,32 @@ func newResourceDelta(
 		delta.Add("", a, b)
 		return delta
 	}
-
-	if ackcompare.HasNilDifference(a.ko.Spec.Description, b.ko.Spec.Description) {
-		delta.Add("Spec.Description", a.ko.Spec.Description, b.ko.Spec.Description)
-	} else if a.ko.Spec.Description != nil && b.ko.Spec.Description != nil {
-		if *a.ko.Spec.Description != *b.ko.Spec.Description {
-			delta.Add("Spec.Description", a.ko.Spec.Description, b.ko.Spec.Description)
+	if len(a.ko.Spec.IngressRules) != len(b.ko.Spec.IngressRules) {
```
knottnt, line 44: nit: could be good to include unit tests that exercise the delta.go behavior.

shabbskagalwala:
seeing this happen as well, would love to have this released 😃 @michaelhtm

PR #337, opened 2026-05-26
feat: add SupportedRegions field to VPCEndpointServiceConfiguration

### LINKED ISSUE #2902: Support of "SupportedRegions" field in VPCEndpointService
**Is your feature request related to a problem?**

Support for the supportedRegions in AWS VPC Endpoint service. 
**Describe the solution you'd like**
Currently the SupportedRegion field is ignored in EC2-controller
**Describe alternatives you've considered**
A description of any alternative solutions or features you've considered.

**I would be happy to contribute back :)**

michaelhtm:
Description of changes:
The field was previously ignored. Users can now specify which regions
service consumers can access the endpoint service from. Uses set.from
to extract Region from SupportedRegionDetail on read, and Add/Remove
semantics on update via ModifyVpcEndpointServiceConfiguration.

Resolves aws-controllers-k8s/community#2902


ncorreia:
Hello, I tested this PR against a real AWS integration environment (built the controller from this branch, applied the updated CRD), and it works. Sharing results in case it helps move the review along.

What I validated:
- The supportedRegions field is accepted on VPCEndpointServiceConfiguration and reconciled to AWS. I set the full set of regions (including the home region) on three live endpoint services and confirmed via describe-vpc-endpoint-service-configurations that all of them were added (SupportedRegions went from 1 to the 18 regions configured).
- The Add path via ModifyVpcEndpointServiceConfiguration works, and the read-back from SupportedRegionDetail populates correctly.

Thanks for putting this together, would love to see it land. 🙏

pkg/resource/vpc_endpoint_service_configuration/hooks.go
```diff
@@ -146,6 +147,71 @@ func (rm *resourceManager) modifyAllowedPrincipals(
 	return nil
 }
 
+func customPreCompare(
```
knottnt, line 150: Q: What's the reason for a custom delta here?

templates/hooks/vpc_endpoint_service_configuration/sdk_update_pre_build_request.go.tpl
```diff
@@ -23,7 +23,15 @@
 		}
 	}
 
+	if delta.DifferentAt("Spec.SupportedRegions") {
```
knottnt, line 26: Since, we're adding custom update behavior it would be good to validate the functionality in an e2e test that checks.

1. That supported regions can be added/removed
2. Updates to other fields applied at the same time are also applied.

knottnt (commented):
Thanks @michaelhtm implementation looks good, but would be good to validate the custom update logic in an e2e test.

pkg/resource/vpc_endpoint_service_configuration/hooks.go
michaelhtm: it allows us to compare supported regions regardless of order.
We can add a helper function in Runtime later to support orderless list comparison

test/e2e/tests/test_vpc_endpoint_service_configuration.py
```diff
@@ -215,4 +216,54 @@ def test_vpc_endpoint_service_configuration_create_delete(self, ec2_client, simp
...
+                "supportedRegions": [REGION],
+            }
+        }
+        k8s.patch_custom_resource(ref, updates)
+        time.sleep(MODIFY_WAIT_AFTER_SECONDS)
+
+        assert k8s.wait_on_condition(ref, "ACK.ResourceSynced", "True", wait_periods=10)
+
+        resource = k8s.get_resource(ref)
+        assert set(resource["spec"]["supportedRegions"]) == {REGION}
```
knottnt, line 257: We should also be asserting that change is reflected in AWS.

knottnt (approved):
/lgtm

PR #352, opened 2026-07-06
Add SourceDestCheckEnabled support to Instance resource

gustavodiaz7722:
Description of changes:
Adds support for managing the EC2 `Instance` `SourceDestCheck` attribute through a new `sourceDestCheckEnabled` spec field. Re-implements the previously closed PR #341.

**Behavior**
- `SourceDestCheck` cannot be set at launch (`RunInstances`), so it is applied post-create via `ModifyInstanceAttribute` once the instance is running. On create with the field set, the value is preserved and the resource is briefly marked `ResourceSynced=False` with the message `Instance is pending update for SourceDestCheckEnabled`.
- Updates (`false` <-> `true`) are handled in `modifyInstanceAttributes`; the current value is read back in `setAdditionalFields`.
- The field is optional. It is configured with `late_initialize` (`skip_incomplete_check`) so the observed value is back-filled into the spec when the user does not set it (or removes it). This keeps `desired` and `latest` in agreement and avoids a phantom delta.
- A guard in `modifyInstanceAttributes` only builds the `SourceDestCheck` input when the desired value is non-nil.

**Why both late_initialize and the nil guard are needed**
The reconciler runs `updateResource` before `lateInitializeResource`. If the field is nil in the desired spec, the standard delta fires and the update path runs first. Without the guard it would call `ModifyInstanceAttribute` with no attribute value and fail (`InvalidParameterCombination: No attributes specified`), returning an error before late-init could restore the value. The guard makes that a no-op so late-init back-fills the observed value on the same reconcile.

**Manual testing observations (live EKS cluster)**
- Create with `sourceDestCheckEnabled: false` -> EC2 `SourceDestCheck=false`, `ResourceSynced=True`.
- Re-enable via patch to `true` -> EC2 `SourceDestCheck=true`, `ResourceSynced=True`.
- Create *without* the field -> late-init populates the spec from the observed value; converges to `ResourceSynced=True` with no update calls.
- Remove the field after it was set (`spec.sourceDestCheckEnabled: null`) -> late-init restores the observed value into the spec; `ResourceSynced=True`, EC2 value preserved, no `ModifyInstanceAttribute` errors and no steady-state reconcile churn.
- Confirmed against the AWS API that a `ModifyInstanceAttribute` request built from a nil value omits the attribute entirely and is rejected with `InvalidParameterCombination: No attributes specified` (it does not silently disable `SourceDestCheck`), which motivated the guard.

**Testing**
- `make build-controller` and `go build ./...` pass.
- E2E test added covering create-with-disabled and re-enable-via-patch.

References: https://github.com/aws-controllers-k8s/community/issues/2883

### DIFF
--- a/pkg/resource/instance/hooks.go
+++ b/pkg/resource/instance/hooks.go
@@ -118,6 +118,8 @@ func (rm *resourceManager) modifyInstanceAttributes(ctx context.Context, delta *
 		input.DisableApiStop = &svcsdktypes.AttributeBooleanValue{Value: desired.ko.Spec.DisableAPIStop}
 	} else if delta.DifferentAt("Spec.SecurityGroupIDs") {
 		input.Groups = aws.ToStringSlice(desired.ko.Spec.SecurityGroupIDs)
+	} else if delta.DifferentAt("Spec.SourceDestCheckEnabled") && desired.ko.Spec.SourceDestCheckEnabled != nil {
+		input.SourceDestCheck = &svcsdktypes.AttributeBooleanValue{Value: desired.ko.Spec.SourceDestCheckEnabled}
 	} else {
 		input = nil
 	}
@@ -157,6 +159,10 @@ func setAdditionalFields(instance svcsdktypes.Instance, ko *v1alpha1.Instance) {
 	ko.Spec.SecurityGroupIDs = []*string{}
 	for _, group := range instance.SecurityGroups {
 		ko.Spec.SecurityGroupIDs = append(ko.Spec.SecurityGroupIDs, group.GroupId)
+	}
+
+	if instance.SourceDestCheck != nil {
+		ko.Spec.SourceDestCheckEnabled = instance.SourceDestCheck
 	}
 
 	if monitoring := instance.Monitoring; monitoring != nil {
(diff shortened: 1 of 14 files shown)


pkg/resource/instance/hooks.go
```diff
@@ -118,6 +118,8 @@ func (rm *resourceManager) modifyInstanceAttributes(ctx context.Context, delta *
 		input.DisableApiStop = &svcsdktypes.AttributeBooleanValue{Value: desired.ko.Spec.DisableAPIStop}
 	} else if delta.DifferentAt("Spec.SecurityGroupIDs") {
 		input.Groups = aws.ToStringSlice(desired.ko.Spec.SecurityGroupIDs)
+	} else if delta.DifferentAt("Spec.SourceDestCheckEnabled") {
+		input.SourceDestCheck = &svcsdktypes.AttributeBooleanValue{Value: desired.ko.Spec.SourceDestCheckEnabled}
 	} else {
```
michaelhtm, line 123: what happens if users have already enabled SourceDestCheck, and upgrade the controller version? Would sending a nil value disable it?

gustavodiaz7722: This is a valid concern, when sending a nil value API will retrun 400.

I have added a fix to ignore field in update and delta when nil. This way we do not send invalid request to AWS.

sapphirew (approved):
/lgtm

PR #357, opened 2026-07-16
feat: sync `IPv6CIDRBlockAssociationSet` while updating VPC state

ramasai1:
https://github.com/aws-controllers-k8s/ec2-controller/pull/330 introduced a change that causes a re-queue of the VPC to delete the default security group rules. This path isn't copying over the IPv6 CIDR block associations, which causes that block to stay empty despite the VPC having a CIDR association present on the console.

Surprisingly, we only see this issue when we're using Amazon provided CIDR blocks, and not when we use IPAM Managed CIDR blocks. Presumably there's a difference between associating a CIDR block with the VPC in both those cases, with the latter returning the CIDR block association immediately after the Create -> ReadOne flow in the runtime, so that the [status is appropriately saved](https://github.com/aws-controllers-k8s/runtime/blob/38bd4c967a0145f81d64505ad8cfe6fdc2d9093c/pkg/runtime/reconciler.go#L860).

### DIFF
--- a/pkg/resource/vpc/hooks.go
+++ b/pkg/resource/vpc/hooks.go
@@ -347,6 +347,7 @@ func (rm *resourceManager) customUpdateVPC(
 		}
 	}
 	updated.ko.Status.CIDRBlockAssociationSet = latest.ko.Status.CIDRBlockAssociationSet
+	updated.ko.Status.IPv6CIDRBlockAssociationSet = latest.ko.Status.IPv6CIDRBlockAssociationSet
 
 	if delta.DifferentAt("Spec.DisallowSecurityGroupDefaultRules") {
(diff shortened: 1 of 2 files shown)


ramasai1:
Hello @michaelhtm - following up on this, thanks.

michaelhtm:
just waiting for tests to pass, and it should be good to merge

ramasai1:
Thank you very much @michaelhtm - does the `release/patch` label mean that an immediate release will be made this with commit, or that it'll be included in the next release?

michaelhtm:
It means a new release PR will be generated after this PR is merged

ramasai1:
Thank you!

PR #359, opened 2026-07-21
fix(security_group): stop perpetual diff on server-normalized rules

gustavodiaz7722:
## Issue

Fixes
- https://github.com/aws-controllers-k8s/community/issues/2822
- https://github.com/aws-controllers-k8s/community/issues/2951

## Description

`SecurityGroup` rules whose spec form differs from the form AWS returns on
`DescribeSecurityGroups` are revoked and re-authorized on **every** reconcile.
Besides the constant churn, a controller crash between the revoke and the
re-authorize leaves the rule missing, degrading any workload that depends on it
(the reporter hit this with CoreDNS node-to-node traffic).

The root cause is that the generated `newResourceDelta` compares
`Spec.IngressRules`/`Spec.EgressRules` with `DeepEqual`, while AWS normalizes
rules server-side on read-back, so the desired and observed forms never compare
equal.

## Fix

Mark `IngressRules`/`EgressRules` with `compare.is_ignored` so the generated
delta skips them, and add a `delta_post_compare` hook (`customPostCompare`) that
compares **canonical copies** of both rule lists and records a delta only for a
genuine difference.

Crucially, the hook works on deep copies and **never mutates** the desired or
latest objects. Those same objects are the merge-patch base the runtime uses to
persist the spec, so the user's `groupRef` (and every other spec field) is left
untouched — the canonical form exists only for the comparison. The
canonicalization is confined to the delta: `syncSGRules` continues to
authorize/revoke the **raw** spec rules exactly as it did before this change, so
a canonicalization defect can only mis-scope which rules a genuine update
touches — it can never change the rule content submitted to AWS.

Canonicalization closes every read-back normalization that drives a spurious
diff:

1. **Reference wrappers & self-references** — `groupRef`/`vpcRef` are spec-only
   wrappers that ACK resolves to concrete ids before the delta and that AWS
   never returns; they are cleared in the copy. A pair is a self-reference when
   `groupID` equals the SG's own id, or when the group identifiers are all
   omitted (the spec shorthand); it is canonicalized by stripping the self
   `groupID` and the server-filled owner `userID` (`groupName` is left
   untouched — see the tradeoff below).
2. **All-protocol (`-1`) rules** — AWS drops the port range for `-1`; we drop
   `fromPort`/`toPort` to match.
3. **Server-filled owner account** — AWS fills `userID` with the owning account
   for same-account grants; we clear it. Cross-account grants (`userID` != owner)
   are preserved so the reference stays intact.
4. **Grant aggregation** — AWS merges grants sharing `(protocol, fromPort,
   toPort)` into one `IpPermission`; we aggregate both sides by that key and sort
   deterministically so ordering never drives a diff.
5. **Protocol notation** — AWS returns well-known protocols by name, so a numeric
   spec value (`"6"`) is mapped to the name AWS returns (`"tcp"`).
6. **CIDR canonicalization** — AWS masks host bits and normalizes IPv6 text
   (`100.68.0.18/18` -> `100.68.0.0/18`); we rewrite CIDRs to the same network
   form.
7. **Read-only peer metadata** — `vpcID`, `peeringStatus` and
   `vpcPeeringConnectionID` are reference metadata EC2 derives from the
   referenced group (populated only for a cross-VPC peer reference) and ignores
   on authorize; the customer cannot set them meaningfully (verified live — a
   bogus spec value is discarded and the derived value returned). We clear them
   in the copy so an AWS-filled cross-VPC read-back never churns against an
   omitted or stale spec value.

The delta assumes `groupID` is resolved by the time it runs; the existing
`referencesResolved` guard already defers rule sync until a `groupRef` resolves,
so the delta never needs to reason about unresolved references.

The fix lives in the manually-maintained `hooks.go`, wired via
`delta_post_compare` and `compare.is_ignored` in `generator.yaml`, so it survives
code-generator regenerations. It builds on the delta-hook idea first proposed in
#325 (thanks @kbujanecki-dt), reworked to compare copies in `delta_post_compare`
so nothing is mutated: genuine rule changes still produce a delta and are
applied, while no meaningful spec data is lost.

### Scope

Only same-account references are normalized for `userID`; cross-account grants
are intentionally left untouched (and preserved).

`groupName` is deliberately **not** canonicalized. AWS never returns a
`groupName` on rule read-back — it resolves any supplied name to a `groupID` and
returns only the id — and a by-name reference is only valid for a group in the
account's default VPC (verified against the EC2 backend and live). Because the
observed state never carries a name, a rule that references a group purely by
`groupName` cannot be reconciled by the delta and will re-diff every reconcile.
Fully supporting it would require a per-read `DescribeSecurityGroups` call to
backfill names into the observed rules (and matching changes to the sync path);
given that by-name references only work in a rare, discouraged default-VPC setup
and `groupID`/`groupRef` is the supported path, that cost is not justified. The
residual churn for the bare-`groupName` case is an accepted tradeoff, and leaving
`groupName` untouched also means a genuine by-name change still surfaces as a
diff rather than being swallowed.

The canonicalization is deliberately confined to delta detection. The sync path
(`syncSGRules`) is left exactly as it was — it matches and authorizes/revokes the
raw spec rules — so this change carries no risk to the code that actually mutates
a user's security group. Improving the sync path itself (see the tradeoff below)
is intentionally out of scope.

### Known tradeoff

Because the sync path operates on the raw spec rules, a genuine update
re-authorizes any rule whose desired spec form differs textually from the form
AWS returns on read-back — an unmasked CIDR (`100.68.0.18/18`), a numeric
protocol (`"6"`), a self-reference by omitted `groupID`, an `icmpv6` rule without
a type/code, or grants written as separate entries that AWS aggregates. Such a
rule is revoked and re-authorized (one revoke + one authorize) even when only a
sibling rule actually changed. This is a one-time action on the edit — not
perpetual churn — and the end state is correct, but it briefly revokes those
rules.

This is a pre-existing property of the sync path and is unchanged by this PR,
which strictly improves delta detection. Moving the sync path to canonical-key or
per-grant matching would narrow the window, but it is intentionally deferred so
the fix stays confined to the delta and the sync path's existing behavior
(delete-before-create, no rollback) is left untouched.

## Testing

**Unit tests** (`pkg/resource/security_group/hooks_test.go`): cover each
canonicalization (`canonicalizeGroupPair`, `canonicalizeRuleList`) and, through
the real `newResourceDelta`, assert every normalization produces **no** diff
while genuine changes (port, protocol, CIDR network, description) still diff.
Additional coverage: the canonicalization runs on copies and never mutates the
source resource; nil-safety of the grant sort; and the inconsistent-input cases
(a foreign `userID` is dropped only when paired with the SG's own `groupID` — an
input AWS itself rejects — while a legitimate cross-account `groupID = peer`
reference is preserved).

**E2E tests** (`test/e2e/tests/test_security_group.py`): each normalization is
exercised in isolation so a regression pinpoints rather than hides behind
another's churn:

- self-reference via omitted `groupID`
- self-reference via `groupRef` (ingress + egress) — also asserts the user's
  `groupRef` survives in the persisted spec after the initial sync and after a
  forced reconcile, with no canonical `groupID` injected
- cross-SG reference to a peer SG — the rule resolves to the **peer** id (not
  self), the server-filled owner `userID` is cleared, and the `groupRef` survives
- all-protocol `-1` port drop
- numeric protocol notation
- IPv4 and IPv6 CIDR canonicalization
- grant aggregation
- multi-grant rule with deterministic sort
- a genuine port change on a self-referencing rule is still applied
- self-reference with a redundant owner `userID` (lossless drop)

Each test creates the SG, verifies the AWS state via the API, then forces a
reconcile with an unrelated (tag) update and asserts the `securityGroupRuleID`s
in `status` do not change — a stable rule id across a forced reconcile is the
definitive signal that no spurious delta re-authorized the rule.

Validated end-to-end against a live EKS cluster: a self-reference (both the
omitted-`groupID` and the `groupRef` form) resolves to `groupID = self`; a
cross-SG `groupRef` referencing a peer resolves to the peer id (not self); the
user's `groupRef` survives in the persisted spec across forced reconciles; and
across multiple forced reconciles the `securityGroupRuleID`s stay stable with
zero `AuthorizeSecurityGroup`/`RevokeSecurityGroup` calls and zero
ingress/egress deltas.

By submitting this pull request, I confirm that my contribution is made under
the terms of the Apache 2.0 license.

### DIFF
--- a/pkg/resource/security_group/hooks.go
+++ b/pkg/resource/security_group/hooks.go
@@ -15,6 +15,9 @@ package security_group
 
 import (
 	"context"
+	"net"
+	"sort"
+	"strconv"
 
 	ackcompare "github.com/aws-controllers-k8s/runtime/pkg/compare"
 	ackcondition "github.com/aws-controllers-k8s/runtime/pkg/condition"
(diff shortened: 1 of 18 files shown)


pkg/resource/security_group/hooks.go
```diff
@@ -129,6 +132,314 @@ func (rm *resourceManager) referencesResolved(
...
+}
+
+// customPreCompare is injected at the top of the generated
+// newResourceDelta (see delta.go) via the `delta_pre_compare` hook in
+// generator.yaml. It rewrites Spec.IngressRules / Spec.EgressRules on both
+// sides into a single canonical form so the subsequent field-by-field
+// DeepEqual does not report spurious diffs that arise purely from how AWS
+// normalises rules on DescribeSecurityGroups read-back.
+//
+// It closes four independent perpetual-diff sources (see
```
knottnt: nit: six issues are listed. 
```suggestion
// It closes six independent perpetual-diff sources (see
```

pkg/resource/security_group/hooks.go
```diff
@@ -129,6 +132,314 @@ func (rm *resourceManager) referencesResolved(
...
+	to := "nil"
+	if rule.ToPort != nil {
+		to = strconv.FormatInt(*rule.ToPort, 10)
+	}
+	return proto + "/" + from + "/" + to
+}
+
+// sortGrants sorts every grant collection of a rule into a deterministic
+// order so that read-back ordering within a rule never drives a diff.
+func sortGrants(rule *svcapitypes.IPPermission) {
```
knottnt, line 473: Should this have a nil guard?

pkg/resource/security_group/hooks.go
```diff
@@ -129,6 +132,314 @@ func (rm *resourceManager) referencesResolved(
...
+) {
+	if pair == nil {
+		return
+	}
+	// A self-reference is either an explicit reference to the SG's own ID
+	// (how AWS returns it, and how ResolveReferences fills a self groupRef)
+	// or a pair that omits all group identifiers (the spec shorthand for
+	// "this SG", since the ID is unknown until AWS assigns it).
+	isSelf := (pair.GroupID != nil && selfID != "" && *pair.GroupID == selfID) ||
+		(pair.GroupID == nil && pair.GroupRef == nil && pair.GroupName == nil)
```
knottnt: Q: why do we need to check GroupRef here? Won't it have been resolved to GroupID at this point?

pkg/resource/security_group/hooks.go
```diff
@@ -129,6 +132,314 @@ func (rm *resourceManager) referencesResolved(
...
+	if isSelf {
+		if selfID != "" {
+			id := selfID
+			pair.GroupID = &id
+		}
+		// GroupRef is a spec-only reference wrapper never present on AWS
+		// read-back; once GroupID is set it is redundant for comparison.
+		pair.GroupRef = nil
```
knottnt: Q: does removing this not impact clearing the resolved reference after the update?

pkg/resource/security_group/hooks.go
```diff
@@ -129,6 +132,314 @@ func (rm *resourceManager) referencesResolved(
...
+		pair.GroupName = nil
+		pair.UserID = nil
```
knottnt: Q: For desired state are these guaranteed to match the self-id? If a user entered an incorrect inputs for these fields it seems like this logic would drop them from the API payload.

pkg/resource/security_group/hooks.go
gustavodiaz7722: Will fix in new revision

gustavodiaz7722: updated the doc comment to "six independent perpetual-diff sources." Thanks for catching it.

pkg/resource/security_group/hooks.go
gustavodiaz7722: Added the guard: sortGrants now returns early on a nil rule, and the per-element comparators use nil-safe sort keys (ipRangeSortKey/ipv6RangeSortKey/prefixListIDSortKey), matching the if r != nil check already in canonicalizeRuleList. Added a unit test that runs canonicalizeRuleList with nil elements in IPRanges/IPv6Ranges/PrefixListIDs (it panicked before, passes now).

pkg/resource/security_group/hooks.go
gustavodiaz7722: After further thought, yes you are correct. During my initial research I was under the impression that self references would not resolve correctly while the securitygroup was creating.

After doing some manual testing, I confirmed your theory that groupID is always resolved, thus we do not need to check groupref.

pkg/resource/security_group/hooks.go
gustavodiaz7722: After testing I have confirmed that the resolved references were in fact being removed.

I will put up a fix to address this so we don't overwrite user references.

gustavodiaz7722: Latest implementation will preserve all user spec fields, including references, by doing a deep copy before canonicalization.

pkg/resource/security_group/hooks.go
```diff
@@ -129,6 +134,390 @@ func (rm *resourceManager) referencesResolved(
...
+	// never drive a diff. GroupID is assumed resolved (referencesResolved
+	// defers sync until it is).
+	pair.GroupRef = nil
+	pair.VPCRef = nil
+
+	// Self-reference: GroupID == selfID, or all group identifiers omitted (the
+	// spec shorthand, since the ID is unknown until AWS assigns it). AWS fills
+	// GroupID/UserID/GroupName on read-back, so canonicalise to {GroupID: selfID}.
+	isSelf := (pair.GroupID != nil && selfID != "" && *pair.GroupID == selfID) ||
+		(pair.GroupID == nil && pair.GroupName == nil)
```
knottnt: Q: If the GroupName and VPCID matches the SG's should it also be considered a self-reference?

pkg/resource/security_group/hooks.go
```diff
@@ -129,6 +134,390 @@ func (rm *resourceManager) referencesResolved(
...
+	if isSelf {
+		if selfID != "" {
+			id := selfID
+			pair.GroupID = &id
+		}
+		pair.GroupName = nil
+		pair.UserID = nil
```
knottnt: If the user enters a value that disagrees with the self-referenced GroupID would this mask that error?

pkg/resource/security_group/hooks_test.go
```diff
@@ -0,0 +1,997 @@
...
+	t.Run("self GroupID + foreign UserID: classified self, foreign UserID dropped", func(t *testing.T) {
+		p := &svcapitypes.UserIDGroupPair{
+			GroupID: aws.String(testSelfID),
+			UserID:  aws.String(testPeerAcctID),
+		}
+		canonicalizeGroupPair(p, testSelfID, testOwnerAcctID)
+		assert.Equal(t, testSelfID, *p.GroupID)
+		assert.Nil(t, p.UserID,
+			"a foreign UserID paired with the SG's own GroupID is nonsensical "+
+				"(AWS rejects it as InvalidGroup.NotFound); dropping it is safe")
```
knottnt: If I'm reading this correctly it sounds like we are silently dropping the bad value if it is the only source of drift. Should we not consider this a delta and send it to AWS?

pkg/resource/security_group/hooks.go
```diff
@@ -129,6 +134,390 @@ func (rm *resourceManager) referencesResolved(
...
+	for _, rule := range rules {
+		if rule == nil {
+			continue
+		}
+		rule.IPProtocol = canonicalizeProtocol(rule.IPProtocol)
+		if rule.IPProtocol != nil {
+			proto := *rule.IPProtocol
+			if _, carriesPorts := portCarryingProtocols[proto]; !carriesPorts {
+				// AWS omits the port range for every protocol outside
+				// {tcp, udp, icmp, icmpv6}: "-1" (all traffic) and any other IP
```
knottnt, line 329: Q: Do we not need to check for -1 here?

pkg/resource/security_group/hooks.go
gustavodiaz7722: Hey so I did some testing against ec2 api. I have confirmed that `VpcId`, `VpcPeeringConnectionId`, and `PeeringStatus` are all read only fields 

EC2 accepts them in the authorize request without validation, silently discards whatever the customer supplies,
and derives the real values from the referenced `GroupId` plus the VPC peering
topology. Only `GroupId` (and `UserId` for cross-account references) actually
determines the reference.


For this reason I believe we should never use these values in the delta, I've updated latest implementation to nil the fields during canonicalization.

pkg/resource/security_group/hooks.go
gustavodiaz7722: -1 would fall under this conditional statement, since it is not in portCarryingProtocols map. I've also added an e2e test to validate -1 protocol specifically.

pkg/resource/security_group/hooks.go
gustavodiaz7722: This is now out of date, but I have updated implementation to take a firm stance on how we handle groupname for canonicalization. 

groupName is deliberately not canonicalized in new implementation. AWS never returns a
groupName on rule read-back — it resolves any supplied name to a groupID and
returns only the id — and a by-name reference is only valid for a group in the
account's default VPC (verified against the EC2 backend and live). Because the
observed state never carries a name, a rule that references a group purely by
groupName cannot be reconciled by the delta and will re-diff every reconcile.
Fully supporting it would require a per-read DescribeSecurityGroups call to
backfill names into the observed rules (and matching changes to the sync path);
given that by-name references only work in a rare, discouraged default-VPC setup
and groupID/groupRef is the supported path, that cost is not justified. The
residual churn for the bare-groupName case is an accepted tradeoff, and leaving
groupName untouched also means a genuine by-name change still surfaces as a
diff rather than being swallowed.

pkg/resource/security_group/hooks_test.go
gustavodiaz7722: This implementation is now out of date. groupName is deliberately not canonicalized in new implementation. See comment above for why we will not support this.

pkg/resource/security_group/hooks.go
knottnt: Okay, I think this makes sense here since the API just silently ignores these values instead of returning an error letting you know you can't actually set the value that we can provide the user as feedback.

pkg/resource/security_group/hooks.go
knottnt: This seems like an acceptable trade-off for now as this essentially falls back to the previous delta behavior.

knottnt (approved):
Thanks @gustavodiaz7722

michaelhtm (approved):

PR #256, opened 2025-03-31
fix: remove printf with no constant

michaelhtm:
Description of changes:
Go 1.24.1 enforcing `fmt.Sprintf("")` to have constant defined.
More info here: https://github.com/golang/go/issues/60529

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #361, opened 2026-07-28
test(vpc): add e2e coverage for ignore-field-drift

sapphirew:
## Description

Adds e2e integration tests exercising the `services.k8s.aws/ignore-field-drift` annotation ([aws-controllers-k8s/runtime#256](https://github.com/aws-controllers-k8s/runtime/pull/256), released in runtime `v0.62.0`) against the EC2 VPC resource.

EC2 is one of the controllers the runtime presubmit (`runtime_presubmit_services`) regenerates and e2e-tests, so this coverage runs on runtime PRs and guards the feature against regressions.

The `IgnoreFieldDrift` feature gate is Alpha and disabled by default, so the tests self-enable it on the deployed controller by patching the deployment's `FEATURE_GATES` env var and waiting for rollout.

That patch rolls the shared controller Deployment, which is a cluster-wide side effect in a suite running 32 xdist workers — a restart mid-run can time out any other worker waiting on `ACK.ResourceSynced`. The gate is therefore turned on **once per run and never turned back off** (check-then-set, so concurrent workers produce a byte-identical pod template and no second rollout; no restore, so a teardown in one worker can't disable the gate underneath a drift test still running in another). Leaving it on is safe because the gate is inert unless a resource carries the annotation, and only this file's resources do.

Follow-up to remove the remaining rollout entirely: set `IgnoreFieldDrift` in `FEATURE_GATES` at controller setup time in [test-infra `scripts/controller-setup.sh`](https://github.com/aws-controllers-k8s/test-infra/blob/main/scripts/controller-setup.sh), alongside the existing `IAMRoleSelector` case, after which this fixture degrades to a no-op check.

## What the tests assert

Four cases, covering the field shapes and drift sources that matter:

**`test_tags_drift_ignored`** — collection field, `spec.tags` ignored:
- the declared tag is applied at create;
- an externally-**added** tag survives — the controller does **not** call `DeleteTags` for drift on the ignored field;
- the resource stays `Synced` even though `spec.tags` differs from the live AWS tag set;
- an edit to `spec.tags` is retained in the CR spec but is **not** pushed to AWS.

**`test_scalar_external_drift_ignored`** — scalar leaf, `spec.enableDNSSupport` ignored. An external `ModifyVpcAttribute` flip survives and the resource stays `Synced`.

**`test_scalar_spec_edit_not_pushed`** — the same scalar, with the edit made while AWS still holds the other value, so declining to push it is observable. Separate test rather than a second half of the one above: the field is boolean, so after an external flip spec and AWS agree and the assertion would be vacuous. It also covers a different code path — `enableDNSSupport` is pushed by `ModifyVpcAttribute` in the custom update path, not by `tags.Sync`.

**`test_declared_tag_value_drift_ignored`** — an externally-changed **value** on a tag key the CR declares. Distinct from an added key: this is the case `tags.Sync` would actively revert.

Note none of these exercise the runtime's path-**prefix** match. Every Delta path here equals its annotation path — `pkg/resource/vpc/delta.go` compares tags as a whole and emits a single `delta.Add("Spec.Tags", ...)`, so `spec.tags` matches by segment equality. Prefix matching needs a Delta subject deeper than the annotation, which none of VPC's ignorable fields produce.

## Testing

Ran locally via `make kind-test SERVICE=ec2` against released runtime `v0.62.0` — no leaked VPCs or CRs (teardown fixture verified clean).

## Files

- `test/e2e/tests/test_vpc_ignore_field_drift.py`
- `test/e2e/resources/vpc_ignore_field_drift.yaml`

## Note on the third commit

`aa14cd9 chore: regenerate apis/v1alpha1/types.go` is unrelated to the tests and is here to get `ec2-verify-code-gen` green.

[code-generator#733](https://github.com/aws-controllers-k8s/code-generator/pull/733) (`d13ed56`, Aug 15) removed the forced `aws.JSONValue` import from `templates/apis/types.go.tpl`. It is not in a release yet — latest is `v0.62.1`, cut two days earlier — but the presubmit regenerates with code-generator `main`, so the committed `types.go` (generated at `v0.62.1`, per `apis/v1alpha1/ack-generate-metadata.yaml`) reads as drifted:

```diff
-	"github.com/aws/aws-sdk-go/aws"
-	_ = &aws.JSONValue{}
```

The commit applies exactly that. `aws-sdk-go` v1 stays in `go.mod` — `pkg/resource/*/hooks.go` still uses it — so nothing else changes. This is not specific to this PR: `main` carries the stale hack, so any ec2-controller PR whose presubmit runs now hits it (`s3-controller` and `iam-controller` are in the same state; `rds-controller` never had the hack, which is why its PRs stay green). Happy to split this into its own PR if you'd rather.

### DIFF
--- a/test/e2e/tests/test_vpc_ignore_field_drift.py
+++ b/test/e2e/tests/test_vpc_ignore_field_drift.py
@@ -1,0 +1,275 @@
+# Copyright Amazon.com Inc. or its affiliates. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License"). You may
+# not use this file except in compliance with the License. A copy of the
+# License is located at
+#
+# 	 http://aws.amazon.com/apache2.0/
+#
+# or in the "license" file accompanying this file. This file is distributed
+# on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+# express or implied. See the License for the specific language governing
+# permissions and limitations under the License.
+
+"""Integration tests for the services.k8s.aws/ignore-field-drift annotation,
+exercised against the EC2 VPC resource and its spec.tags field.
+
+This is the runtime feature under aws-controllers-k8s/runtime#256. The
+IgnoreFieldDrift feature gate is Alpha and disabled by default, so the test
+enables it on the deployed controller for the duration of the module and
+restores the prior value afterwards. EC2 is one of the controllers the runtime
+presubmit regenerates and e2e-tests, so this coverage runs on runtime PRs.
+"""
+
+import logging
+import time
+
+import boto3
+import pytest
+
+from acktest import tags
+from acktest.k8s import condition
+from acktest.k8s import resource as k8s
+from acktest.resources import random_suffix_name
+from e2e import service_marker, CRD_GROUP, CRD_VERSION, load_ec2_resource
+from e2e.replacement_values import REPLACEMENT_VALUES
+from e2e.tests.helper import EC2Validator
+
+RESOURCE_PLURAL = "vpcs"
+PRIMARY_CIDR_DEFAULT = "10.0.0.0/16"
+
+CREATE_WAIT_AFTER_SECONDS = 15
+MODIFY_WAIT_AFTER_SECONDS = 20
+DELETE_WAIT_AFTER_SECONDS = 10
+
+# Controller deployment coordinates in the kind test cluster (see
+# test-infra/scripts/controller-setup.sh and run-e2e-tests.sh).
+CONTROLLER_NAMESPACE = "ack-system"
+CONTROLLER_DEPLOYMENT = "ack-ec2-controller"
+CONTROLLER_CONTAINER = "controller"
+FEATURE_GATE = "IgnoreFieldDrift"
+# Generous window for the new pod to roll out and take over reconciliation.
+ROLLOUT_WAIT_SECONDS = 120
+
+
+def _apps_client():
+    # Build the AppsV1Api against acktest's configured ApiClient (which points
+    # at the kind cluster). A bare AppsV1Api() would default to localhost:80.
+    from kubernetes import client as k8s_client
+    return k8s_client.AppsV1Api(k8s._get_k8s_api_client())
+
+
+def _get_feature_gates_env() -> str:
+    """Returns the current value of the FEATURE_GATES env var on the controller
+    container, or "" if it is unset."""
+    dep = _apps_client().read_namespaced_deployment(
+        CONTROLLER_DEPLOYMENT, CONTROLLER_NAMESPACE,
+    )
+    for c in dep.spec.template.spec.containers:
+        if c.name != CONTROLLER_CONTAINER:
+            continue
+        for e in (c.env or []):
+            if e.name == "FEATURE_GATES":
+                return e.value or ""
+    return ""
+
+
+def _set_feature_gates_env(value: str):
+    """Patches the FEATURE_GATES env var on the controller container and waits
+    for the rollout to complete. The controller wires this env var into its
+    --feature-gates flag (see the controller deployment manifest)."""
+    body = {
+        "spec": {
+            "template": {
+                "spec": {
+                    "containers": [
+                        {"name": CONTROLLER_CONTAINER,
+                         "env": [{"name": "FEATURE_GATES", "value": value}]},
+                    ]
+                }
+            }
+        }
+    }
+    _apps_client().patch_namespaced_deployment(
+        CONTROLLER_DEPLOYMENT, CONTROLLER_NAMESPACE, body,
+    )
+    _wait_for_rollout()
+
+
+def _merge_gate(existing: str, gate: str, enabled: bool) -> str:
+    """Returns a FEATURE_GATES string with `gate` set to `enabled`, preserving
+    any other gates already present."""
+    pairs = {}
+    for part in filter(None, (p.strip() for p in existing.split(","))):
+        if "=" in part:
+            k, v = part.split("=", 1)
+            pairs[k.strip()] = v.strip()
+    pairs[gate] = "true" if enabled else "false"
+    return ",".join(f"{k}={v}" for k, v in pairs.items())
+
+
+def _wait_for_rollout():
+    """Blocks until the controller deployment reports all replicas updated and
+    available for the current generation."""
+    client = _apps_client()
+    deadline = time.time() + ROLLOUT_WAIT_SECONDS
+    while time.time() < deadline:
+        dep = client.read_namespaced_deployment(
+            CONTROLLER_DEPLOYMENT, CONTROLLER_NAMESPACE,
+        )
+        spec_replicas = dep.spec.replicas or 1
+        status = dep.status
+        if (status.observed_generation is not None
+                and status.observed_generation >= dep.metadata.generation
+                and (status.updated_replicas or 0) >= spec_replicas
+                and (status.available_replicas or 0) >= spec_replicas
+                and (status.unavailable_replicas or 0) == 0):
+            # Give the fresh pod a moment to acquire leadership / start reconciling.
+            time.sleep(5)
+            return
+        time.sleep(3)
+    raise AssertionError(
+        f"controller deployment {CONTROLLER_DEPLOYMENT} did not roll out within "
+        f"{ROLLOUT_WAIT_SECONDS}s after toggling the {FEATURE_GATE} feature gate"
+    )
+
+
+@pytest.fixture(scope="module")
+def ec2_client():
+    from acktest.aws.identity import get_region
+    return boto3.client("ec2", get_region())
+
+
+@pytest.fixture(scope="module")
+def ignore_field_drift_enabled():
+    """Enables the IgnoreFieldDrift feature gate on the controller for the
+    duration of the module, then restores the prior FEATURE_GATES value."""
+    original = _get_feature_gates_env()
+    _set_feature_gates_env(_merge_gate(original, FEATURE_GATE, True))
+    yield
+    # Restore exactly what was there before (which may be "").
+    _set_feature_gates_env(original)
+
+
+@pytest.fixture
+def ignore_field_drift_vpc(request):
+    """A VPC annotated with ignore-field-drift on spec.tags, carrying one
+    declared tag (team=payments) as the create-time baseline."""
+    resource_name = random_suffix_name("vpc-ifd-test", 24)
+    replacements = REPLACEMENT_VALUES.copy()
+    replacements["VPC_NAME"] = resource_name
+    replacements["CIDR_BLOCK"] = PRIMARY_CIDR_DEFAULT
+    replacements["ENABLE_DNS_SUPPORT"] = "False"
+    replacements["ENABLE_DNS_HOSTNAMES"] = "False"
+    replacements["ENABLE_NETWORK_ADDRESS_USAGE_METRICS"] = "False"
+    replacements["DISALLOW_DEFAULT_SECURITY_GROUP_RULE"] = "False"
+    replacements["TAG_KEY"] = "team"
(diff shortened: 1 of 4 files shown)


sapphirew:
Heads up: I pushed a temporary commit (`41ed2d8`, titled **DO NOT MERGE**) that adds a `go.mod` replace pointing the ACK runtime at my fork branch (runtime PR #256, the `IgnoreFieldDrift` gate).

**Why:** the `ec2-kind-e2e` check builds the controller from the checked-in `go.mod`, which pins the released runtime `v0.61.0`. That runtime doesn't know the `IgnoreFieldDrift` gate, and gate parsing is strict — the controller `os.Exit(1)`s at startup on `FEATURE_GATES=IgnoreFieldDrift=true`, so the new tests fail with "did not roll out". The replace lets CI build a controller that recognizes the gate and actually exercise the tests once.

**Merge plan:** runtime #256 merges & releases first → ec2-controller bumps to a runtime version containing the gate → I revert `41ed2d8` → this PR merges. The replace commit must not merge as-is.

gustavodiaz7722:
> [!NOTE]
> This comment was generated by a Gen AI code review agent and reviewed before posting.

### MUST FIX

1. **[Category: tests]**
   - File: `test/e2e/tests/test_vpc_ignore_field_drift.py` (`test_nested_field_under_ignored_parent`)
   - Issue: The test claims to exercise the runtime's path-**prefix** match with a Delta path of `Spec.Tags.N.Value` under an ignored `spec.tags`. No such path exists. `pkg/resource/vpc/delta.go` compares tags as a whole via `convertToOrderedACKTags` + `MapStringStringEqual` and emits exactly one difference, `delta.Add("Spec.Tags", ...)`. So `Path.ContainsFold("spec.tags")` matches by exact segment equality here — the same branch `test_tags_drift_ignored` already covers. Two independent confirmations: `runtime/pkg/compare/path_test.go` asserts `Path("Spec.KMSKeyID").ContainsFold("spec.kmsKeyID.extra")` is **False**, i.e. `Contains` requires the *ignored* path to be a prefix of the *Delta* path, so prefix matching needs a Delta path **deeper** than the annotation; and `isValidFieldPath` in `ignore_field_drift.go` rejects indices outright, so `Spec.Tags.N.Value` is not even expressible as an annotation.
   - Fix: Either (a) pick a field whose generated delta genuinely emits a nested path — a nested struct member where code-gen produces a `compareX` helper with a `Spec.Parent.Child` subject — and annotate the parent; or (b) drop the prefix-match claim. Option (b) is fine: tag-**value** drift on a declared key is a materially different case from a **new** externally-added key (it is the one that `tags.Sync` would actively revert), so keep the test but rename it to `test_declared_tag_value_drift_ignored` and correct the docstring plus the PR description. Do not leave a test asserting coverage of a branch it never reaches.

2. **[Category: tests]**
   - File: `test/e2e/tests/test_vpc_ignore_field_drift.py` (`test_scalar_field_drift_ignored`, final block)
   - Issue: The "edit retained but not pushed" assertions are vacuous. The external actor has already set the live attribute to `True` via `modify_vpc_attribute`; the test then patches `spec.enableDNSSupport` to `True` and asserts the live attribute is `True` and the spec is `True`. Both hold whether or not the edit was pushed — spec and AWS agree, so there is no delta to suppress and nothing being tested. Contrast `test_tags_drift_ignored`, which patches to `team=changed` while AWS holds `payments`, and is therefore sound.
   - Fix: Reorder so the edit is made against a *differing* live value. Create with `enableDNSSupport: False` (the controller applies it at create), patch the spec to `True`, assert the live attribute is still `False` and `spec.enableDNSSupport is True` — that is the real retain/not-pushed assertion. Then do the external-flip half. Because the field is boolean there is no third value, so ordering is the only way to make both halves non-vacuous.

3. **[Category: tests]**
   - File: `test/e2e/tests/test_vpc_ignore_field_drift.py` (`ignore_field_drift_vpc` teardown)
   - Issue: `except: pass` around `assert deleted` silently discards teardown failure, so a leaked VPC or CR passes unnoticed — which directly undercuts the PR description's claim that teardown was verified clean. The bare form also swallows `KeyboardInterrupt` and `SystemExit`.
   - Fix: Remove the try/except and let `assert deleted` fail the teardown, or at minimum narrow to `except Exception:`, log, and re-raise. Separately, move the `delete_tags` cleanup of the out-of-band `external` tag out of the test body into the fixture teardown (or a `finally`), so a mid-test assertion failure does not leave it behind.

test/e2e/tests/test_vpc_ignore_field_drift.py
```diff
@@ -0,0 +1,456 @@
...
+        [{"ignore_paths": "spec.tags"}],
+        indirect=True,
+    )
+    def test_nested_field_under_ignored_parent(
+        self, ec2_client, ignore_field_drift_enabled, ignore_field_drift_vpc,
+    ):
+        """Drift on a nested child of an ignored parent: the declared tag's
+        value (Delta path Spec.Tags.N.Value) changes externally while only the
+        parent spec.tags is ignored. The runtime must match this by path prefix,
+        so the child drift is ignored too."""
```
gustavodiaz7722: _Gen AI generated comment._

- Issue: The test claims to exercise the runtime's path-**prefix** match with a Delta path of `Spec.Tags.N.Value` under an ignored `spec.tags`. No such path exists. `pkg/resource/vpc/delta.go` compares tags as a whole via `convertToOrderedACKTags` + `MapStringStringEqual` and emits exactly one difference, `delta.Add("Spec.Tags", ...)`. So `Path.ContainsFold("spec.tags")` matches by exact segment equality here — the same branch `test_tags_drift_ignored` already covers. Two independent confirmations: `runtime/pkg/compare/path_test.go` asserts `Path("Spec.KMSKeyID").ContainsFold("spec.kmsKeyID.extra")` is **False**, i.e. `Contains` requires the *ignored* path to be a prefix of the *Delta* path, so prefix matching needs a Delta path **deeper** than the annotation; and `isValidFieldPath` in `ignore_field_drift.go` rejects indices outright, so `Spec.Tags.N.Value` is not even expressible as an annotation.
- Fix: Either (a) pick a field whose generated delta genuinely emits a nested path — a nested struct member where code-gen produces a `compareX` helper with a `Spec.Parent.Child` subject — and annotate the parent; or (b) drop the prefix-match claim. Option (b) is fine: tag-**value** drift on a declared key is a materially different case from a **new** externally-added key (it is the one that `tags.Sync` would actively revert), so keep the test but rename it to `test_declared_tag_value_drift_ignored` and correct the docstring plus the PR description. Do not leave a test asserting coverage of a branch it never reaches.

test/e2e/tests/test_vpc_ignore_field_drift.py
```diff
@@ -0,0 +1,456 @@
...
+        )
+
+        # Editing the ignored scalar in the spec is retained but NOT pushed to
+        # AWS: patch to a new value and confirm the live attribute is unchanged.
+        k8s.patch_custom_resource(ref, {"spec": {"enableDNSSupport": True}})
+        time.sleep(MODIFY_WAIT_AFTER_SECONDS)
+
+        assert _dns_support_enabled(ec2_client, vpc_id) is True
+        latest = k8s.get_resource(ref)
+        assert latest["spec"].get("enableDNSSupport") is True
```
gustavodiaz7722: _Gen AI generated comment._

- Issue: The "edit retained but not pushed" assertions are vacuous. The external actor has already set the live attribute to `True` via `modify_vpc_attribute`; the test then patches `spec.enableDNSSupport` to `True` and asserts the live attribute is `True` and the spec is `True`. Both hold whether or not the edit was pushed — spec and AWS agree, so there is no delta to suppress and nothing being tested. Contrast `test_tags_drift_ignored`, which patches to `team=changed` while AWS holds `payments`, and is therefore sound.
- Fix: Reorder so the edit is made against a *differing* live value. Create with `enableDNSSupport: False` (the controller applies it at create), patch the spec to `True`, assert the live attribute is still `False` and `spec.enableDNSSupport is True` — that is the real retain/not-pushed assertion. Then do the external-flip half. Because the field is boolean there is no third value, so ordering is the only way to make both halves non-vacuous.

test/e2e/tests/test_vpc_ignore_field_drift.py
```diff
@@ -0,0 +1,456 @@
...
+    assert cr is not None
+    assert k8s.get_resource_exists(ref)
+
+    yield (ref, cr)
+
+    try:
+        _, deleted = k8s.delete_custom_resource(ref, 3, 10)
+        assert deleted
+    except:
+        pass
```
gustavodiaz7722: _Gen AI generated comment._

- Issue: `except: pass` around `assert deleted` silently discards teardown failure, so a leaked VPC or CR passes unnoticed — which directly undercuts the PR description's claim that teardown was verified clean. The bare form also swallows `KeyboardInterrupt` and `SystemExit`.
- Fix: Remove the try/except and let `assert deleted` fail the teardown, or at minimum narrow to `except Exception:`, log, and re-raise. Separately, move the `delete_tags` cleanup of the out-of-band `external` tag out of the test body into the fixture teardown (or a `finally`), so a mid-test assertion failure does not leave it behind.

test/e2e/tests/test_vpc_ignore_field_drift.py
sapphirew: Confirmed, and taken option (b). `pkg/resource/vpc/delta.go:143-146` emits a single `delta.Add("Spec.Tags", ...)` via `convertToOrderedACKTags` + `MapStringStringEqual`, and `Path.Contains` (`runtime/pkg/compare/path.go:63-77`) bails when the subject has more segments than the path, so `spec.tags` against `Spec.Tags` is segment equality — the branch `test_tags_drift_ignored` already reaches.

Renamed to `test_declared_tag_value_drift_ignored` in 13e3586. The docstring now states what it covers (a changed value on a *declared* key, which `tags.Sync` would actively revert, unlike an added key) and explicitly what it does not (prefix matching needs a Delta subject deeper than the annotation; per-element paths are not expressible since `isValidFieldPath` rejects indices). Corrected the class docstring, which repeated the same claim, and the PR description.

test/e2e/tests/test_vpc_ignore_field_drift.py
sapphirew: Confirmed vacuous — the external flip sets the live value to `True` before the spec is patched to `True`, so the assertion holds either way.

One correction to the proposed fix: reordering within a single test does not resolve it, because the field is boolean. After `create False` → patch spec `True`, asserting the live value is still `False` is sound; but then flipping live to `True` makes spec and AWS agree, so the external-drift half asserts nothing. Whichever half runs second finds no delta to suppress, and there is no third value to escape with.

Split into two tests in 13e3586, each with its own VPC:

- `test_scalar_external_drift_ignored` — create `False`, external flip to `True`, assert it survives and stays Synced.
- `test_scalar_spec_edit_not_pushed` — create `False`, patch spec to `True`, assert the live value is still `False` and the spec retains `True`.

The edit case also earns a separate test on its own merits: `enableDNSSupport` reaches AWS through `ModifyVpcAttribute` in the custom update path rather than `tags.Sync`, so suppression has to hold for that path too.

test/e2e/tests/test_vpc_ignore_field_drift.py
sapphirew: Both points fixed in 13e3586. The `try`/`except: pass` is gone, so a failed delete fails the teardown, and the out-of-band `external` tag cleanup moved into a `finally` — as a trailing statement it was skipped whenever an assertion above it failed.

One addition: the delete budget goes from 30s to 60s in the same change. `delete_custom_resource(ref, 3, 10)` gives a VPC CR 30s to clear its finalizer, and ec2's own `test_vpc.py::TestVpc::test_crud_tags` failed on exactly `assert deleted is True` in a run on this PR earlier this week. Asserting on the old budget would have traded a silent leak for a flake, so the assertion now runs against a budget that reflects observed deletion times.

PR #368, opened 2026-08-26
fix(managed_prefix_list): stop perpetual diff on entry order

gustavodiaz7722:
Issue #, if available: none filed yet.

Description of changes:

`ManagedPrefixList` never reaches `Synced=True` after an entries update. The resource settles on `Synced=Unknown` and stays there, while AWS holds the correct entries.

The generated `newResourceDelta` compares `Spec.Entries` with an order-sensitive `DeepEqual`:

```go
if !equality.Semantic.Equalities.DeepEqual(a.ko.Spec.Entries, b.ko.Spec.Entries) {
    delta.Add("Spec.Entries", a.ko.Spec.Entries, b.ko.Spec.Entries)
}
```

`sdkFind` fills `Spec.Entries` from `GetManagedPrefixListEntries`, which returns entries in an order AWS chooses. Once that order differs from the declared one, `Spec.Entries` differs on every reconcile, and no update can resolve it: `computeEntriesToAdd` and `computeEntriesToRemove` key off CIDR, so both return empty.

`customUpdateManagedPrefixList` issued a request regardless, because `PrefixListName` was set unconditionally and so always satisfied the "only call if there are actual changes" guard. That request carries `CurrentVersion` with no entry modification, which AWS rejects:

```
InvalidParameterCombination: You cannot include a version unless your request
includes a modification to the entries of the prefix list.
```

`customUpdate` returns the error, `ensureConditions` records `Synced=Unknown`, and the next reconcile repeats it.

### Fix

Three changes:

1. Mark `Entries` with `compare.is_ignored` and add a `delta_post_compare` hook that compares the entries as a CIDR-keyed set, matching how the add and remove sets are already computed. This is the treatment `SecurityGroup` rules received in #359.
2. Set `PrefixListName` only when `Spec.Name` differs, so a reconcile with nothing to change no longer builds a request.
3. Set `CurrentVersion` only when the request modifies entries. Per the EC2 API, entry additions and removals create a new version while a name change does not, so a version on a request that modifies no entries is what AWS rejects.

The first change fixes the loop. The other two prevent a request that cannot succeed from being built at all.

The comparison lives in the manually maintained `hooks.go`, wired through `generator.yaml`, so it survives regeneration. Regenerated with code-generator `v0.62.1`, matching the `build_hash` already recorded in `apis/v1alpha1/ack-generate-metadata.yaml`; `api_directory_checksum` is unchanged, so no API types moved.

### Testing

**Unit** (`pkg/resource/managed_prefix_list/hooks_test.go`): nine cases asserted through the real `newResourceDelta`, so the `generator.yaml` wiring is covered alongside the comparison. Same entries in a different order produce no delta; added, removed, and replaced entries, and a changed or added description on the same CIDR, all still produce one. Plus nil-safety for `customPostCompare`.

**Live cluster** (EKS, 60s resync). The scenario from `test_update_entries`: create with three entries, then add a fourth. AWS returned `10.0.0.0/24  10.0.1.0/24  10.0.2.0/24  192.168.1.0/24` against a declared order of `10.0.0.0/24  10.0.1.0/24  192.168.1.0/24  10.0.2.0/24`, so the order divergence was present in both arms.

| | before | after |
|---|---|---|
| CR after update | `Synced=Unknown` for 40 polls | `Synced=True` in ~35s |
| `ACK.Recoverable` | `InvalidParameterCombination` | absent |
| AWS state | `modify-complete`, version 2 | `modify-complete`, version 2 |
| `ModifyManagedPrefixList` errors in log | repeating | none |

Held `Synced=True` across more than a full resync period with the AWS version pinned at 2, confirming no residual diff drives repeat modifications. A tags-only update, the scenario from `test_update_tags`, converges immediately and leaves the version at 2.

### Note on the e2e tests

`test_update_entries` and `test_update_tags` pass on `main` today, so this is not a visible CI failure. They pass because `customUpdate` returns `desired`, which the runtime also uses as the base of the status merge patch, making the patch empty and leaving the CR on the stale `Synced=True` from creation. The 400 loop happens either way; only the reporting is suppressed. aws-controllers-k8s/runtime#265 fixes that aliasing, at which point both tests fail against this resource. Merging this first keeps that change green.

### DIFF
--- a/pkg/resource/managed_prefix_list/hooks.go
+++ b/pkg/resource/managed_prefix_list/hooks.go
@@ -49,6 +49,29 @@ func updateTagSpecificationsInCreateRequest(r *resource,
 	}
 }
 
+// customPostCompare compares Spec.Entries as a set keyed by CIDR.
+//
+// GetManagedPrefixListEntries returns entries in an order AWS chooses, which
+// need not match the order declared in the Spec. An order-sensitive comparison
+// therefore reports a difference that no update can ever resolve: the entries
+// to add and remove are computed from CIDR-keyed maps, so both come back empty
+// and the next reconcile sees the same difference again.
+func customPostCompare(
+	delta *ackcompare.Delta,
+	a *resource,
+	b *resource,
+) {
+	if a == nil || b == nil || a.ko == nil || b.ko == nil {
+		return
+	}
+	if !ackcompare.MapStringStringEqual(
+		buildEntriesMap(a.ko.Spec.Entries),
+		buildEntriesMap(b.ko.Spec.Entries),
+	) {
+		delta.Add("Spec.Entries", a.ko.Spec.Entries, b.ko.Spec.Entries)
+	}
+}
+
 // customUpdateManagedPrefixList provides custom logic for updating ManagedPrefixList
 func (rm *resourceManager) customUpdateManagedPrefixList(
 	ctx context.Context,
@@ -75,8 +98,12 @@ func (rm *resourceManager) customUpdateManagedPrefixList(
 	input := &svcsdk.ModifyManagedPrefixListInput{}
 	input.PrefixListId = latest.ko.Status.ID
 
-	// Always set the Name field from desired state
-	input.PrefixListName = desired.ko.Spec.Name
+	// Only set the Name when it actually changed. Setting it unconditionally
+	// defeats the "are there changes?" guard below, so a reconcile with nothing
+	// to change still issues a request.
+	if delta.DifferentAt("Spec.Name") {
+		input.PrefixListName = desired.ko.Spec.Name
+	}
 
 	if delta.DifferentAt("Spec.MaxEntries") {
 		if desired.ko.Spec.MaxEntries != nil {
@@ -103,8 +130,11 @@ func (rm *resourceManager) customUpdateManagedPrefixList(
 		}
 	}
 
-	// Set current version for optimistic locking (required by AWS for any modification)
-	if latest.ko.Status.Version != nil {
+	// Set current version for optimistic locking. Only entry modifications
+	// create a new version, so AWS rejects CurrentVersion on a request that
+	// carries none.
+	if (len(input.AddEntries) > 0 || len(input.RemoveEntries) > 0) &&
+		latest.ko.Status.Version != nil {
 		input.CurrentVersion = latest.ko.Status.Version
 	}
 
--- a/apis/v1alpha1/ack-generate-metadata.yaml
+++ b/apis/v1alpha1/ack-generate-metadata.yaml
@@ -1,9 +1,9 @@
 ack_generate_info:
-  build_date: "2026-08-26T02:49:09Z"
-  build_hash: 65d45b2e6c9efd6aca20e0d36826d1e18e4ba2b7
+  build_date: "2026-08-26T18:34:44Z"
+  build_hash: d4aaca5e1dd6acfd81094936dc825ad30d2bbb39
   go_version: go1.26.0
-  version: v0.62.1
-api_directory_checksum: 55717efb3996c41b42b7978b2f20551c7ad5cdcc
+  version: v0.62.1-4-gd4aaca5
+api_directory_checksum: a80968686fdeff40b93f27e9672dfad4d0cfa109
 api_version: v1alpha1
 aws_sdk_go_version: v1.41.2
 generator_config_info:
--- a/apis/v1alpha1/ack-generate-metadata.yaml
+++ b/apis/v1alpha1/ack-generate-metadata.yaml
@@ -1,13 +1,13 @@
 ack_generate_info:
-  build_date: "2026-08-26T18:34:44Z"
+  build_date: "2026-08-26T19:34:36Z"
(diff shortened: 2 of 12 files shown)


pkg/resource/managed_prefix_list/hooks.go
```diff
@@ -75,8 +98,12 @@ func (rm *resourceManager) customUpdateManagedPrefixList(
 	input := &svcsdk.ModifyManagedPrefixListInput{}
 	input.PrefixListId = latest.ko.Status.ID
 
-	// Always set the Name field from desired state
-	input.PrefixListName = desired.ko.Spec.Name
+	// Only set the Name when it actually changed. Setting it unconditionally
+	// defeats the "are there changes?" guard below, so a reconcile with nothing
+	// to change still issues a request.
+	if delta.DifferentAt("Spec.Name") {
+		input.PrefixListName = desired.ko.Spec.Name
```
sapphirew, line 105: Making `PrefixListName` conditional means the "are there actual changes?" guard below can now silently swallow a real diff.

`PrefixListName` was previously the only unconditionally-set field, so it always satisfied that guard. Now a reconcile whose only diff is `Spec.AddressFamily` builds an empty request, fails the guard, and returns `desired, nil` — no API call, no error. The CR still reports `Synced=True` (`synced.when` matches `state=create-complete`), so the diff never clears and the user gets no signal. Before this change the same path produced a visible `InvalidParameterCombination`.

https://github.com/aws-controllers-k8s/ec2-controller/blob/9d9743f3d82855bb22a2015994c9098e31845198/pkg/resource/managed_prefix_list/hooks.go#L140-L144

`AddressFamily` is a required Spec field with no immutability marker, and `ModifyManagedPrefixListInput` has no `AddressFamily` parameter, so it is unreachable by any update:

https://github.com/aws-controllers-k8s/ec2-controller/blob/9d9743f3d82855bb22a2015994c9098e31845198/apis/v1alpha1/managed_prefix_list.go#L31-L33

This is the concern raised on #359 ("we are silently dropping the bad value if it is the only source of drift. Should we not consider this a delta and send it to AWS?").

Not blocking the entries fix. Two options: mark `AddressFamily` `is_immutable: true` in `generator.yaml` so the runtime raises a terminal error, or return a terminal error when the delta is non-empty but no request field was set.

🤖 Generated with [Claude Code](https://claude.ai/code)

sapphirew (approved):
no blocking comment
/lgtm

pkg/resource/managed_prefix_list/hooks.go
gustavodiaz7722: Confirmed on all three points, and fixed in c430777.

`ModifyManagedPrefixListInput` has no `AddressFamily` member — zero occurrences in the whole operation file — so the field is unreachable by any update and always has been. What my change altered is only the signal. The unconditional `PrefixListName` and `CurrentVersion` used to push out a request that AWS rejected with `InvalidParameterCombination`, which was wrong about the cause but visible. Gating both traded that for silence, and `synced.when` only inspects `Status.State`, so the CR reports `Synced=True` while the drift persists. A misleading error is bad; silence is worse.

Took the first of your two options: `AddressFamily` is now `is_immutable: true`. Two reasons for preferring it to a terminal error in `customUpdate`.

It matches what this repo already does. `VPCEndpoint.ServiceRegion` and `ServiceNetworkARN` carry the same marking with the same rationale — absent from the Modify input — so the comment I added is deliberately worded like theirs. And it acts earlier: `is_immutable` emits a CEL `self == oldSelf` rule, so the API server rejects the change at admission with `Value is immutable once set` before a reconcile ever runs. The user finds out at `kubectl patch`, not by reading conditions afterwards.

The blanket "delta is non-empty but no request field was set" variant I considered and rejected as dead code. Walking the compared fields:

- `Tags` returns via `!delta.DifferentExcept("Spec.Tags")` before reaching the guard
- `Name` and `MaxEntries` both map to request fields, and `MaxEntries` is required so it cannot be nil
- the CIDR-keyed entries delta only fires when the maps differ, which means an add or a remove is always non-empty — a description change on an existing CIDR is caught by `computeEntriesToAdd`'s `currentDesc != desc`

`AddressFamily` was the only unreachable field, so after marking it there is no remaining path to an empty request. I'd rather add the catch-all when a second such field actually appears than carry a branch that cannot currently fire.

Verified on a live cluster:

| | result |
|---|---|
| `addressFamily` IPv4 → IPv6 | rejected: `spec.addressFamily: Invalid value: "string": Value is immutable once set` |
| unrelated tag change | accepted |
| re-applying the same `addressFamily` | accepted, no change |

The third case matters: `self == oldSelf` is a transition rule, so it only evaluates on update. Creates are unconstrained, existing resources are untouched until someone tries to change the field, and idempotent re-applies of an unchanged manifest still pass.

On #359's open question — whether drift that cannot be sent should be surfaced rather than dropped — this is the narrow version of that answer: for a field AWS has no parameter for, immutability is the honest encoding, since no amount of surfacing makes it reconcilable. The broader case, where a value *could* be sent but is being silently skipped, is still worth its own look and isn't addressed here.

Sorry for the re-review this costs — pushing dismissed the lgtm.

PR #372, opened 2026-09-10
Propagate Spec.Tags to volumes and network interfaces at instance launch

### LINKED ISSUE #2954: [ec2-controller] Expose RunInstancesInput.TagSpecifications in Instance CR spec to support tag-based SCP enforcement
## Describe the feature

The EC2 controller's `generator.yaml` currently sets `RunInstancesInput.TagSpecifications` to `ignore` ([ref: generator.yaml#L110](https://github.com/aws-controllers-k8s/ec2-controller/blob/main/generator.yaml#L110C2-L110C42)). This prevents users from specifying `TagSpecifications` in the `Instance` CR, meaning tags cannot be applied to sub-resources (e.g., EBS volumes, network interfaces) at instance launch time.

## Why is this needed?

Many organizations enforce **tag-based Service Control Policies (SCPs)** via AWS Organizations that require specific tags to be present at resource creation time using `aws:RequestTag` conditions. For example:

```json
{
  "Effect": "Deny",
  "Action": "ec2:RunInstances",
  "Resource": "arn:aws:ec2:*:*:volume/*",
  "Condition": {
    "Null": {
      "aws:RequestTag/kubernetes.io/created-for/pvc/name": "true"
    }
  }
}
```

Since ACK does not include `TagSpecifications` in the `RunInstances` API call, the SCP denies the operation with:

> `UnauthorizedOperation: ec2:RunInstances on resource: volume/* with an explicit deny in a service control policy`

**Post-creation tagging via `CreateTags` does NOT satisfy SCP conditions**, because SCP evaluates `aws:RequestTag` only at resource creation time.

## Proposed Solution

Remove the `ignore` directive on `RunInstancesInput.TagSpecifications` in `generator.yaml` and expose it as a field in the `Instance` CR spec. This would allow users to pass `TagSpecifications` at instance launch time, enabling:

- Tag-based SCP compliance for EBS volumes, ENIs, etc. created with the instance
- Cost allocation tags applied at creation
- Compliance with organizational tagging policies

## Environment

- **EC2 Controller Version:** v1.15.0
- **Kubernetes Version:** 1.28+

## Additional Context

- Reference to ignored field: https://github.com/aws-controllers-k8s/ec2-controller/blob/main/generator.yaml#L110C2-L110C42
- Related (high-level design): https://github.com/aws-controllers-k8s/community/issues/63

gobos12:
Issue #, if available: Resolves [community#2954](https://github.com/aws-controllers-k8s/community/issues/2954)

Description of changes:
This change extends the Instance create hook (`updateTagSpecificationsInCreateRequest`) so that Spec.Tags are applied not just to the instance but also to the EBS volumes and network interfaces created during `RunInstances`, via the launch call's `TagSpecifications`. This lets tag-based SCPs that gate on `aws:RequestTag` (e.g. requiring a tag on `volume/*`) pass at creation time, which post-creation `CreateTags` cannot satisfy. Sub-resource tags are create-only — the instance's tags continue to be reconciled on update, but volume and network-interface tags are set once at launch and not tracked afterward (no per-sub-resource ID tracking required). Adds a unit test for the hook and an e2e test asserting the launch tags land on the sub-resources at create and that a later `Spec.Tags` update reconciles the instance without touching the volume/ENI tags. 


config/controller/kustomization.yaml
```diff
@@ -6,4 +6,4 @@ kind: Kustomization
 images:
 - name: controller
   newName: public.ecr.aws/aws-controllers-k8s/ec2-controller
-  newTag: 1.21.0
+  newTag: 0.0.0-non-release-version
```
gustavodiaz7722: Don't touch release tags, these are auto generated by prow job for cutting new releases.

helm/Chart.yaml
```diff
@@ -1,8 +1,8 @@
 apiVersion: v1
 name: ec2-chart
 description: A Helm chart for the ACK service controller for Amazon Elastic Cloud Compute (EC2)
-version: 1.21.0
-appVersion: 1.21.0
+version: 0.0.0-non-release-version
```
gustavodiaz7722: Don't touch release tags, these are auto generated by prow job for cutting new releases.

helm/values.yaml
```diff
@@ -4,7 +4,7 @@
 
 image:
   repository: public.ecr.aws/aws-controllers-k8s/ec2-controller
-  tag: 1.21.0
+  tag: 0.0.0-non-release-version
```
gustavodiaz7722: Don't touch release tags, these are auto generated by prow job for cutting new releases.

helm/templates/NOTES.txt
```diff
@@ -1,5 +1,5 @@
 {{ .Chart.Name }} has been installed.
-This chart deploys "public.ecr.aws/aws-controllers-k8s/ec2-controller:1.21.0".
+This chart deploys "public.ecr.aws/aws-controllers-k8s/ec2-controller:0.0.0-non-release-version".
```
gustavodiaz7722: Don't touch release tags, these are auto generated by prow job for cutting new releases.

gustavodiaz7722:
- **File:** `pkg/resource/instance/hooks.go` — `launchTagResourceTypes`, `updateTagSpecificationsInCreateRequest`
- **Issue:** the code comment states *"If a launch does not create a listed resource type,
  RunInstances errors; that validation is left to the EC2 API."* Deferring to the API here turns
  a working configuration into a create failure.

  AWS documents the general rule on
  [TagSpecification](https://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_TagSpecification.html):
  tagging a resource type the action does not support returns an error. The launch-specific form
  is explicit on
  [LaunchTemplateTagSpecificationRequest](https://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_LaunchTemplateTagSpecificationRequest.html):
  if the instance does not include the resource type you specify, the launch fails — and it names
  volumes as the counterexample, since not every instance type includes one. The observed ENI
  error is `You cannot specify tags for network interfaces if there are no network interfaces
  being created by the request`
  ([report](https://teamcity-support.jetbrains.com/hc/en-us/community/posts/11195133835922-AWS-EC2-LaunchTemplate-Fails-With-Attached-Network-Interface-amazon-sync-action-RunWithTags)).
  *(Sources rephrased for compliance with licensing restrictions.)*

- **Reachability — verified end to end.** The attach-existing-ENI path is fully plumbed:
  - `spec.networkInterfaces[].networkInterfaceID` is user-settable —
    `config/crd/bases/ec2.services.k8s.aws_instances.yaml:462`
  - it is copied into the launch request —
    `pkg/resource/instance/sdk.go:1693`, inside `newCreateRequestPayload` (declared at line 1292):
    `f27elem.NetworkInterfaceId = f27iter.NetworkInterfaceID`

- **Blast radius is larger than `spec.tags` users.** The hook reads `Spec.Tags` *after* ACK has
  injected its default tags, and those defaults are on by default:
  - `runtime/pkg/runtime/reconciler.go:851-853` calls `EnsureTags(ctx, desired, ...)` immediately
    before `rm.Create(ctx, desired)`, with an in-code comment saying it re-ensures tags
    specifically so they are present for the create
  - `runtime/pkg/config/config.go:77` — `defaultResourceTags` carries
    `services.k8s.aws/controller-version` and `services.k8s.aws/namespace`
  - `ec2-controller/helm/values.yaml:122-129` — the chart adds `app.kubernetes.io/managed-by` and
    `kro.run/kro-version` on top

  So `Spec.Tags` is effectively never empty at create, the `if r.ko.Spec.Tags == nil` early return
  never fires in a normal deployment, and all three tag specifications are emitted on **every**
  Instance create. Any Instance CR that attaches a pre-existing ENI now fails to launch whether or
  not the user set a single tag.

- **Fix:** derive the resource-type list from the request instead of hardcoding it. Include
  `network-interface` only when the launch will create at least one ENI — that is, when
  `Spec.NetworkInterfaces` is empty, or when at least one entry has `NetworkInterfaceID == nil`
  (a mixed list still creates an ENI, so it stays eligible). For `volume`, either apply the same
  treatment or, if you intend to keep it unconditional, say so in the PR description and
  acknowledge the instance-store-backed-AMI failure mode that AWS's own docs call out.

blackdog0403:
Thanks for picking this up, [@gobos12](https://github.com/aws-controllers-k8s/ec2-controller/pull/372#) — this addresses the issue I filed in [community#2954](https://github.com/aws-controllers-k8s/community/issues/2954).

Just wanted to add some context relevant to [@gustavodiaz7722](https://github.com/aws-controllers-k8s/ec2-controller/pull/372#)'s review.

The use case driving this issue does involve the attach-existing-ENI pattern (spec.networkInterfaces[].networkInterfaceID) in some environments. So the dynamic approach Gustavo suggested — including each resource type only when the launch actually creates it — aligns well with how ACK is used in practice.

For reference, Terraform's AWS provider takes a similar approach: it builds the TagSpecifications list dynamically based on which sub-resources the RunInstances call will actually create, rather than hardcoding all resource types unconditionally.

